PR #61, opened 2014-09-01
Dumper: Multiple live snapshots

JanTvrdik:


TomasVotruba:
What needs to be done here?

JanTvrdik:
We need to figure whether we want this functionality and then pick one of MANY implementation/approaches we came up with.

TomasVotruba:
@JanTvrdik What exactly is the feature and what issue does it solve?

I miss that in PR description and I don't understand following comments. Could you update the description please?

dg:
This is just idea with very low priority.

JanTvrdik:
> What exactly is the feature and what issue does it solve?

Live dumps (see Dumper::LIVE) are amazing technology invented by @dg for Tracy – such dumps are lazy-rendered using javascript (only visible parts of the dump are rendered). 

1) It greatly improves rendering speed of large dumps.
2) It greatly reduces the overall size (as in bytes) of the dump 
2a) because the serialization format does not contains the HTML tags
2b) because each object instance is dumped only once
3) As a side effect the dumps are often deeper without decreasing performance.

Currently this technology is used for all dumps in Bluescreen. This PR is trying to make this technology available for dumps outside of bluescreen.

EdaCZ:
@JanTvrdik @dg Can I help with something to move this forward? Or is there any problem?

TomasVotruba:
@JanTvrdik Nice points. Some after/before screenshots would help to understand that in practice.

TomasVotruba:
@EdaCZ I'm not sure what the blocker is, but I see only this one:
T> @dg This is just idea with very low priority.

 @dg could you explain please? If I understand @JanTvrdik correctly:

> @JanTvrdik  Currently this technology is used for all dumps in Bluescreen. This PR is trying to make this technology available for dumps outside of bluescreen.

this is **just Bluescreen : Tracy feature compatibility**.

JanTvrdik:
> Can I help with something to move this forward? Or is there any problem?

There is no problem, it just takes time to resolve this properly. The implementation was done in multiple ways and many other ways how this can be implemented have been discussed in theory (read carefully all related PR comments + their code to understand this better). What's missing is an overall analysis of all possibilities and impacts.

> Some after/before screenshots would help to understand that in practice.

No, they would not help at all. The dumps look (almost)  identical. This is not a visual improvement.

dg:
Implemented in v2.6

PR #77, opened 2014-11-03
XDebug backtrace can be disabled

ovlach:
Because Helpers::fixStack can generate little different backtrace then PHP backtrace and some loggers works only with PHP backtrace (for example function named *unknown* in  Helpers::fixStack...) we need option to disable xdebug backtrace


dg:
@ovlach is there any other difference besides `*unknown*`?

dg:
@ovlach I think that better is to remove differences between Xdebug and Tracy, than add new option. Can you provide some example, when Tracy with / without Xdebug generates different code?

dg:
closing, no response

PR #81, opened 2014-11-19
Debugger Bar: is loaded in separated HTTP request & AJAX

dg:


src/Tracy/Bar.php
```diff
@@ -17,12 +17,36 @@
...
+	public function register()
+	{
+		if (isset($_GET['_tracy_bar'])) {
+			$this->dispatch($_GET['_tracy_bar']);
+			exit;
+		}
+
+		$this->id = md5(uniqid(TRUE));
+		$this->handle = fopen(($this->directory ?: ini_get('upload_tmp_dir')) . '/bar.' . $this->id, 'w');
+		flock($this->handle, LOCK_EX);
```
milo: Why lock? What about exception when already registered?

src/Tracy/Bar.php
```diff
@@ -105,9 +129,49 @@ public function render()
...
+
+
+	private function dispatch($command)
+	{
+		if (preg_match('#^content.(\w+)$#', $command, $m)) {
+			$directory = $this->directory ?: ini_get('upload_tmp_dir');
+			$file = ($this->directory ?: ini_get('upload_tmp_dir')) . '/bar.' . $m[1];
+			header('Content-Type: text/javascript');
+			if ($content = @file_get_contents($file)) {
+				echo 'Tracy.Debug.newContent(' . json_encode(Helpers::fixEncoding($content)) . ');';
```
milo: Isn't here some way, how to print it "pretty"? Hard to check in the source.

src/Tracy/Bar.php
```diff
@@ -105,9 +129,49 @@ public function render()
...
+				unlink($file);
+			}
+			return;
+		}
+
+		switch ($command) {
```
milo: What about some caching headers for these statics?

milo:
One bug found... when opening panel into window, floating panel stays and get
```
var winPanel = win.Tracy.Debug.getPanel(this.id); <-- TypeError: win.Tracy is undefined
```
in [here](https://github.com/dg/tracy/blob/7be1711d2c97d7e606ab030f650a8ea95d00f66f/src/Tracy/templates/bar.js#L152). Otherwise works great!

src/Tracy/Bar.php
dg: I wrote this 3 years ago, so I don't now, but it is important.

dg: I already know! The other part (currently not published) of this PR, support for AJAX, requires locking, but here it is useless

src/Tracy/Bar.php
dg: Do you mean to load pure HTML?

src/Tracy/Bar.php
dg: Good point.

src/Tracy/Bar.php
milo: Ideally. When I developed panel on current Tracy, sometimes I made mistake and it was hard to find where in such one-line code. I helped myself to dump it by var_dump() but this would be more developer friendly.

milo:
Two notes before I forget:
- it store bar into `D:\` when run from CLI
- dumps are not colored nor clickable when simple `<?php Tracy\Debugger::enable(); dump($_SERVER)`

PR #91, opened 2015-01-05
[WIP] Added simple time tracking panel

stekycz:
Shows all timer results in time panel for easier time debugging. See http://forum.nette.org/cs/21752-vysledky-mereni-casu-v-debbuger-baru

**Questions**
- <del>Default listener for logging times of `Application` and `Presenter` should be in [application](https://github.com/nette/application/tree/master/src/Bridges/ApplicationTracy) I guess, right?</del> Not related, see https://github.com/nette/tracy/pull/91#issuecomment-68796507
- Should `disable` option exist in NEON configuration (using `CompilerExtension`)?
- What about time panel ID? Is it ok to name it as I did or should I use generated name in method `addPanel` as ID to get the panel?
- Are all used information needed in table?
- Is there some way (or should be) to create some "tree" of times? For example [`Application::onStartup`](https://github.com/nette/application/blob/master/src/Application/Application.php#L87) could be collapsed hiding all times generated during this event (in all its listeners).


JanTvrdik:
I don't see a point in built-in measuring `Application` or `Presenter`. It will tell you nothing useful.

stekycz:
I think it will. There is a lot of things you can measure - action, handles, beforeRender, render, afterRender, shutdown etc. It could tell in which part of the application life cycle could be a bottle neck.

It is also taken from mentioned topic, see http://forum.nette.org/cs/21752-vysledky-mereni-casu-v-debbuger-baru#p148559

JanTvrdik:
@stekycz You should use a profiler for this. Measuring it with `Tracy` will slow down application and provide very little of actually useful data.

stekycz:
@JanTvrdik Yes, I use profiles when it is needed. I probably understand what you mean. This feature is probably for extra extension only for those who would like to use it.

src/Tracy/Debugger.php
```diff
@@ -379,7 +379,7 @@ public static function getBar()
 	{
 		if (!self::$bar) {
 			self::$bar = new Bar;
-			self::$bar->addPanel(new DefaultBarPanel('time'));
+			self::$bar->addPanel(new DefaultBarPanel('time'), 'time');
```
JanTvrdik: You should probably use `__CLASS__ . ':time'` here.

src/Tracy/Debugger.php
```diff
@@ -470,6 +470,13 @@ public static function timer($name = NULL)
 		$now = microtime(TRUE);
 		$delta = isset($time[$name]) ? $now - $time[$name] : 0;
 		$time[$name] = $now;
+		if (!self::$productionMode) {
+			static $panel;
+			if (!$panel) {
+				$panel = self::getBar()->getPanel('time');
+			}
+			$panel->data[] = array('name' => $name, 'time' => $now, 'delta' => $delta);
+		}
```
JanTvrdik: I would consider doing just 

~~~php
if (!self::$productionMode) {
    self::getBar()->getPanel(__CLASS_ . ':time')->data[] = array('name' => $name, 'time' => $now, 'delta' => $delta);
}
~~~

Your solution suffers from performance asymmetry because first `Debugger::timer()`  call would be slower than the second one.


----

Also only the `$time` is irrelevant. Only `$delta` matter. Therefore you should only do this when `$delta` is not `0`.

src/Tracy/Debugger.php
stekycz: Yes, it could be better. I could also use [default logic](https://github.com/nette/tracy/blob/master/src/Tracy/Bar.php#L38) without need of own ID. However I am not sure what is better.

src/Tracy/Debugger.php
stekycz: I agree with zero `$delta`. It makes sense to me. However usage of `$delta` only has no advantage against to `bd(Debugger::time())`. Maybe the default label is `bd(Debugger::time('foo'), 'foo')`. The information about timestamp of timer "event" and the time since last timer usage is the advantage.

The performance improvement you suggest is super micro optimization. I cannot measure any difference in visible `ms` units. It also affects development environment only so it does not matter in my opinion.

JanTvrdik: > The information about timestamp of timer "event" and the time since last timer usage is the advantage.

Ok, you convinced me. Keep the `$now` info. We we'll see how useful it is in practice.

> The performance improvement 

Improvement? I made it intentionally slower by removing your performance improvement (static cache for `$panel`). I agree that **your** improvement was super micro optimization, so I removed it.

src/Tracy/Debugger.php
JanTvrdik: Custom ID is better.

src/Tracy/Debugger.php
stekycz: Ups, my bad :-) I agree with you now. It was a focus lost.

src/Tracy/templates/bar.time.panel.phtml
```diff
@@ -0,0 +1,38 @@
...
+
+namespace Tracy;
+
+use Tracy;
+
+if (empty($data)) {
+	return;
+}
+?>
+<h1>Times: <?php echo count($data) ?></h1>
```
JanTvrdik: I don't see how the `count($data)` is useful.

JanTvrdik:
> Is there some way (or should be) to create some "tree" of times?

That would be cool, however it is not possible because the calls may look like this:

~~~php
Debugger::timer('a');
Debugger::timer('b');
Debugger::timer('b');
Debugger::timer('b');
Debugger::timer('c');
Debugger::timer('a');
Debugger::timer('c');
~~~

stekycz:
There could be some optional parameter with meaning of depth. If not specified it takes current internal depth. For example:
```php
Debugger::timer('a', -1);
Debugger::timer('b');
Debugger::timer('b');
Debugger::timer('b');
Debugger::timer('c');
Debugger::timer('a', +1);
Debugger::timer('c');
```

fprochazka:
I'd suggest finishing the current implementation first. Improvements can be done separately :)

JanTvrdik:
> There could be some optional parameter with meaning of depth. 

No, there really could not be some magical optional parameter which nobody would use.

----

The biggest problem with the current implementation (assuming all may suggestions are accepted) is that the last two columns have confusing names. One is "since last" and the other is "since last with same name" but named just "delta". But "since last" is also *a delta*. I have no idea for better column names however.

dg:
Panel `time` was removed, but this can be nice extension.

stekycz:
@dg Do you mean extra extension or extension included in Nette? Current implementation (without time panel) could be updated to use this idea.

dg:
It is developed and maintained by you, so your extension.

stekycz:
Understood

dg:
When it will be mature and will prove that it is useful and you will be willing to do it, we can merge it to Tracy.

stekycz:
We will see :-)

PR #92, opened 2015-01-19
Logger: Added REMOTE_ADDR

o5:
I would like to know a remote address, when I see into log files like log/access.log. Of course, I can go to logs from nginx/apache server and find them but this is a faster way.


dg:
This is BC break for logging format.

fprochazka:
:-1: 

If you wanna have custom logs, use [kdyby/monolog](https://github.com/kdyby/monolog). You can register custom processor, that will add whatever you need to the log.

o5:
@fprochazka thanks for sharing, I'll try it.

PR #93, opened 2015-01-20
Added bridge for Nette Framework

dg:
see nette/bootstrap#4

Closes #58

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,12 +18,14 @@
 		"php": ">=5.3.1"
 	},
 	"require-dev": {
+		"nette/di": "~2.3",
 		"nette/tester": "~1.0"
 	},
 	"autoload": {
-		"classmap": ["src/Tracy"],
+		"classmap": ["src"],
 		"files": ["src/shortcuts.php"]
 	},
+	"minimum-stability": "dev",
 	"extra": {
 		"branch-alias": {
 			"dev-master": "2.3-dev"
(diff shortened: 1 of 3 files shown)


src/Bridges/Nette/Bridge.php
```diff
@@ -0,0 +1,61 @@
...
+							. BlueScreen::highlightFile($item['args'][0], $m[1])
+					);
+				} elseif ($item = Helpers::findTrace($e->getTrace(), 'Nette\Neon\Decoder::decode')) {
+					return array(
+						'tab' => 'NEON',
+						'panel' => BlueScreen::highlightPhp($item['args'][0], $m[1])
+					);
+				}
+			}
+		});
```
JanTvrdik: Why is this not in latte/latte and nette/neon?

dg: There are things that I do not know where to put. I hate it.

dg: Perhaps I'll leave it in the nette/bootstrap and solve it sometime in the future.

JanTvrdik: > There are things that I do not know where to put

Could you write a list of those things to https://github.com/nette/bootstrap/issues/4? Maybe someone will have a good idea.

dg: This file and rest of NetteExtension https://github.com/dg/nette-bootstrap/commit/c5793aa494dce77b06d61ca438b5f56e5040dc0d#diff-d4b5afc3710c208bf26e8d1fcf72f288L278

PR #95, opened 2015-02-04
Don't set error_log, log_errors is set to FALSE anyway.

pascalhofmann:


dg:
Why?

pascalhofmann:
I think nothing is ever logged to the error_log, because log_errors is set to false some lines below.

dg:
Good point.

dg:
Merged e4beb992367f8889d2394068e8f66fac881d843c

PR #98, opened 2015-02-08
Removed memory tab

dg:


hrach:
:+1: I understand the icon as a temporary, so I don't care.

vysinsky:
:+1:

src/Tracy/templates/bar.statistics.panel.phtml
```diff
@@ -0,0 +1,41 @@
...
+	<td>Execution time</td>
+	<td><b><?php echo number_format((microtime(TRUE) - Debugger::$time) * 1000, 1, '.', ' ') ?></b> ms</td>
+</tr>
+<tr>
+	<td>The peak of allocated memory</td>
+	<td><b><?php echo number_format(memory_get_peak_usage() / 1000000, 2, '.', ' ') ?></b> MB</td>
+</tr>
+<tr>
+	<td>Included files</td>
+	<td><b><?php echo count(get_included_files()), '</b> (', number_format(array_sum(array_map('filesize', get_included_files())) / 1000000, 2, '.', ' ') ?> MB)</td>
```
fprochazka: i don't think the filesize is neccesary

dg: Only execution time is neccesary.

enumag:
:+1:

JanTvrdik:
How is IP address and PHP version related to statistics? Why not make Tracy logo clickable instead and write environment info there instead of the title attribute?

dg:
Renamed to 'info'

JanTvrdik:
Will you simplify the Tracy logo title?

hrach:
I don't see added value in number of classes etc, at least strip funcs. It's hard to read the rest.

fprochazka:
`Types + functions: 344 + 2` would be nicer imho

Majkl578:
@fprochazka: Trait is not a type.

Issue #97, opened 2015-02-07
New Bar icons


medhi:
Are you open to change the bar icons? I would suggest to use Fontawesome font to generate bigger icons to look good on retina.

My suggestions:

Execution time: http://fontawesome.io/icon/clock-o/
Memory: http://fontawesome.io/icon/bar-chart/
Routes: http://fontawesome.io/icon/road/
Database: http://fontawesome.io/icon/database/
Identity: http://fontawesome.io/icon/user/

I suppose, 32x32px should be enough for retina displays.

License allows use icons without attribution, but i would suggest to mention it somewhere.

dg:
Do you have any preview, how it will look?

medhi:
![snimek obrazovky 2015-02-07 v 22 08 37](https://cloud.githubusercontent.com/assets/522848/6093988/f078cab2-af15-11e4-9be9-024222133dac.png)

dg:
Thank you!

enumag:
@medhi I like that. Especially the routing icon. :-)

dg:
As with the redesign of the operating system, the 3rd party addons will now look ugly.

fprochazka:
:+1: lovely

solcik:
:+1:

milo:
Looks nice but colors needed. Isn't purpose of icons to make searching things faster?

dg:
For comparsion:

![2015-02-07_22-50-47](https://cloud.githubusercontent.com/assets/194960/6094102/f072d610-af1b-11e4-8d45-7cec1cd8fe86.png)

medhi:
I'll try make coloured preview.

dg:
My font is not as bold as your, probably cause I am using Windows, can you place here screenshot of current design?

Majkl578:
@dg: Easy solution, CSS grayscale filter. http://davidwalsh.name/convert-images-grayscale

dg:
Ad icons: why to generate bitmaps from fonts? SVG is fine.

soundake:
:+1:

kacer:
:-1: I prefer coloured icons. It is easier to distinguish them.

fprochazka:
It would be nice to have icons gray and colored on hover (which can be done by css filter).

TomPastyrik:
[...]

PR #99, opened 2015-02-09
vector icons 

dg:
Vector icons. Colors can be changed.
See #97


hrach:
I have no idea what is the second icon. Also I don't like the box for session. I liked more the dollar sign icon for dumps, but the eye is ok.

dg:
Second icon is Dependency Injection container. And feel free to suggest better icons.

hrach:
- Session: what about something like folder? http://fontawesome.io/icon/folder-open/
- DIC: this http://fontawesome.io/icon/tasks? it remindes me defition list.

dg:
Btw, we are absolutely not limited to fontawesome.io

dg:
Zoomed:

hrach:
What about removing the "opacity overlay" for previous render and fixing the the bg glitch on the right side? :o)

medhi:
For session, I suggest a chair.
tracy-kreslo

dg:
Good idea :-)

hrach:
I think, that we can strip "session" label <= chair is awesome! :-)

fprochazka:
Chair is lovely :)

dg:
Can you send link to chair?

medhi:
Home made :)

https://gist.github.com/medhi/cab98cdfd9be83964d4b#file-chair-svg

medhi:
And here I tried to create a container with some injection :) Alternative could be injection itself or just some type of container bin.
tracy-kreslo-di

medhi:
:+1:

hrach:
I liked more the blue graph, this icons seems more like signal strenght than graph.

fprochazka:
The dollar sign was nicer. The DIC is interesting, I guess we won't be needing title for that one :) saved space ftw.

medhi:
Maybe better without the gradient on the bar graph.

@hrach Quite, but signal icon has always the bars in the incremental order. But I agree, that previous icon was more clear.

dg:
ad blue vs orange: I like more blue too (obviously, my favourite color), but Bar usually will contain these items and Route is blue too:

untitled-2

ad signal strengh: maybe last column can be lower

ad dollar sign: it is not variables, it is dumps

medhi:
Yes, blue is better. Make the route icon darker: #3A404A

dg:
What about this:
untitled-2

dg:
I am not still sure with pink database ;-)

medhi:
Coloured graph looks little bit messy. Database should has been violet. I don't know, from what I remember it from, it so long. Old times :)

dg:
New icons are merged, you can check them in sandbox.

dg:
One more thing: route icon looks good on retina display, but on other displays it is heavy A. Maybe we should find anoter icon. Maybe something with arrow? Like https://github.com/nette/tracy/issues/97#issuecomment-73431224

medhi:
https://www.iconfinder.com/icons/200504/arrows_directional_signs_two_icon#size=128
https://www.iconfinder.com/icons/200507/arrows_directional_divide_opus_signs_two_icon#size=128

dg:
Btw this https://www.iconfinder.com/icons/200508/arrows_directional_signs_two_way_icon#size=128 looks in low resolution like coitus ;-)

medhi:
Ok, we found an icon looking like a coitus. At least one achievement of our night discussion :)

dg:
:-))

dg:
untitled-2

eugeene:
neat!

hrach:
I would like to see time by default, please! :-)

eugeene:
I can see the time by default

hrach:
Yes, sorry, I just saw tweet, which says it is elastic, which probably means it's responsive :-)

dg:
This screenshot is taken with `.tracy-label {display: none;}` and big zoom. I wanted to emphasize new icons.

But even so is the new Tracy shorter.

medhi:
Red "logout" icon and one icon with gradient? I really don't like it, but still better than old bitmap icons :)

dg:
Does it look like logout?

medhi:
Reversed logout :)

hrach:
It looks like exit. https://www.google.com/search?q=exit+icon&source=lnms&tbm=isch&sa=X&ei=qefZVLSeKNPkaMq6goAC&ved=0CAgQ_AUoAQ&biw=1600&bih=763

dg:
It is simply the same icon as the older bitmap one. We can change it.

mancze:
I just really hate that gradient on graph icon. From where did it come? Every other icon is flat and this is THE one which has gradient.

I don't like the exit icon neither but at least it is just vectorized version of the old icon so it makes some sense.

dg:
All icons are flat, some of them has more colors or shades.

vojtech-dobes:
To be honest, I love the gradient.

mrtnzlml:
This is nice! Now it's time to redesign also panels right? Not just bar... (-:

Just one thing. There was Nette version on Tracy logo hover in previous bar. It was great. But in the new one, there is just Tracy version. Is that mistake or not?

george-oakling:
@mrtnzlml - Tracy is indipendent on Nette, versioning is displayed only for it - its logical...it shloudnt be problem to implement an extension to show Nette version, but for me, it would be nonsense. Tracy version is what matters on Tracy toolbar :)

eugeene:
@george-oakling it is, but, in my opinion, it is better to show Tracy version when used stand-alone, but show Nette version when used /w framework.

dg:
@mrtnzlml it will be visible in next RC (https://github.com/nette/bootstrap/commit/6ef8507bccdd3ebe0fdd2aefed17a3ea53015a62)

PR #101, opened 2015-02-20
TracyExtension: added options logDirectory and logSeverity

JanTvrdik:
Not tested, but hopefully it will work =)

### DIFF
--- a/src/Bridges/Nette/TracyExtension.php
+++ b/src/Bridges/Nette/TracyExtension.php
@@ -19,6 +19,7 @@ class TracyExtension extends Nette\DI\CompilerExtension
 {
 	public $defaults = array(
 		'email' => NULL,
+		'logSeverity' => NULL,
 		'editor' => NULL,
 		'browser' => NULL,
 		'errorTemplate' => NULL,


fprochazka:
The logDirectory would be a bit schizophrenic now.

src/Bridges/Nette/TracyExtension.php
```diff
@@ -19,6 +19,8 @@ class TracyExtension extends Nette\DI\CompilerExtension
 {
 	public $defaults = array(
 		'email' => NULL,
+		'logDirectory' => NULL,
+		'logSeverity' => 0,
```
dg: Why not NULL?

JanTvrdik:
Probably a noob question, but how do you write `E_ALL` in NEON? I tried

~~~
tracy:
	logSeverity: E_ALL
~~~

and

~~~
tracy:
	logSeverity: ::constant('E_ALL')
~~~

The first generates

~~~
Tracy\Debugger::$logSeverity = 'E_ALL';
~~~

and the second on generates

~~~
Tracy\Debugger::$logSeverity = Nette\PhpGenerator\Helpers::createObject('Nette\DI\Statement', array(
	"\x00Nette\\DI\\Statement\x00entity" => '::constant',
	'arguments' => array('E_ALL'),
));
~~~

src/Bridges/Nette/TracyExtension.php
JanTvrdik: I know nothing about writing extension and thought that the value should correspond to default value of `Debugger::$logSeverity`.

dg: To rewrite what user sets in boostrap.php is bad. https://github.com/JanTvrdik/nette-tracy/blob/f47bdf5dccfdc26dbb61cf7c4cf1f128ef2ffd7e/src/Bridges/Nette/TracyExtension.php#L71

dg:
Like `::E_ALL`, but TracyExtension is not ready for this. On the other hand, it to use constant is not enough, something like `E_WARNING | E_NOTICE` should be supported to.

`createObject(...)` is bug, I'll try to fix it.

src/Bridges/Nette/TracyExtension.php
JanTvrdik: What if we change `if ($value !== NULL)` to `if ($value !== $this->defaults[$key])`?

JanTvrdik:
@fprochazka Could you elaborate on why would allowing `logSeverity` in `config.neon` cause any problems?

fprochazka:
You should set the log directory as soon as possible - which means in bootstrap - you may have errors during compilation and you need them logged. Also, having two different directories for logging in aplication is not something I would call smart. And if you have no reason to change the directory, because you have to set it in bootstrap, and you would just copy the same value, that is only a space for errors IMHO - because you're configuring the same thing from two separate places.

fprochazka:
If the tracy extension were first and you would set it also in `loadConfiguration` directly to `Tracy\Debugger`, expecting no errors in nette ever to occure, than it would might make sense. Now it doesn't for me.

dg:
loadConfiguration in first extension is tooooo late.

src/Bridges/Nette/TracyExtension.php
dg: And change `email` etc. to `''`. Seems like unnecessary complication.

fprochazka:
Exactly :)

src/Bridges/Nette/TracyExtension.php
JanTvrdik: Maybe for Tracy, but (generally speaking) it allows you to generate optimal code which in some cases may reduces number of included files.

JanTvrdik:
> E_WARNING | E_NOTICE should be supported to

I think that in most situation you want to log everything (E_ALL) or nothing (0).

dg:
Probably, but what about minor situations ;-) But it can be easily solved is extensions. 

The question is, whether to configure this option is not the same issue as logDirectory. (This also applies to half the other options.)

PR #107, opened 2015-04-01
PHP 7: exceptionHandler may accept EngineException as well

JanTvrdik:


Majkl578:
It's not enough imho. Debugger::_exceptionHandler still accepts only \Exception.
Also it's pretty stupid that it has to be untyped, but only way would be some class aliasing on <7 I guess.

JanTvrdik:
Note: this only fixes compatibility with PHP 7, there are many more things which will need to be reconsidered and changed due to introduction of `EngineException` and other error-related changes in PHP 7.

JanTvrdik:
@dg In case you would decide to merge it, don't =) Another change is being considered https://wiki.php.net/rfc/throwable-interface

dg:
Throwable seems much better than BaseException.

PR #108, opened 2015-04-05
Bar: info bar show HHVM version if is running on HHVM

Vrtak-CZ:


hrach:
Better not to add hhvm key if it's not running under hhvm.

Majkl578:
@hrach: :+1: 
Same for PHP.

Vrtak-CZ:
@hrach @Majkl578 is in `array_filter` https://github.com/nette/tracy/blob/master/src/Tracy/assets/Bar/info.panel.phtml#L14

hrach:
@Vrtak-CZ oh, sorry, obviously 3 lines around isn't enough :)

Majkl578:
@Vrtak-CZ: PHP will not be filtered.

PR #109, opened 2015-04-06
added fromEmail option to set From header

fifco:
Useful for hosting where From email headers in php mail() are
restricted to concrete ones. Can be configured through neon.

- feature
- documentation - needed
- BC break - no

### DIFF
--- a/src/Tracy/Debugger.php
+++ b/src/Tracy/Debugger.php
@@ -132,7 +132,7 @@ class Debugger
 	 * @param  string  sender email; in 'From' header when sending email
 	 * @return void
 	 */
-	public static function enable($mode = NULL, $logDirectory = NULL, $email = NULL, $fromEmail = NULL)
+	public static function enable($mode = NULL, $logDirectory = NULL, $email = NULL)
 	{
 		self::$time = isset($_SERVER['REQUEST_TIME_FLOAT']) ? $_SERVER['REQUEST_TIME_FLOAT'] : microtime(TRUE);
 		error_reporting(E_ALL | E_STRICT);
@@ -144,9 +144,6 @@ class Debugger
 		// logging configuration
 		if ($email !== NULL) {
 			self::$email = $email;
-		}
-		if ($fromEmail !== NULL) {
-			self::$fromEmail = $fromEmail;
 		}
 		if ($logDirectory !== NULL) {
 			self::$logDirectory = $logDirectory;
@@ -407,7 +404,7 @@ class Debugger
 	public static function getLogger()
 	{
 		if (!self::$logger) {
-			self::$logger = new Logger(self::$logDirectory, self::$email, self::getBlueScreen(), self::$fromEmail);
+			self::$logger = new Logger(self::$logDirectory, self::$email, self::getBlueScreen());
 			self::$logger->directory = & self::$logDirectory; // back compatiblity
 			self::$logger->email = & self::$email;
 			self::$logger->fromEmail = & self::$fromEmail;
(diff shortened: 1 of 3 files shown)


src/Tracy/Debugger.php
```diff
@@ -400,9 +407,10 @@ public static function setLogger(ILogger $logger)
 	public static function getLogger()
 	{
 		if (!self::$logger) {
-			self::$logger = new Logger(self::$logDirectory, self::$email, self::getBlueScreen());
+			self::$logger = new Logger(self::$logDirectory, self::$email, self::getBlueScreen(), self::$fromEmail);
```
fprochazka: I don't like this new property and I see no reason for the debugger to be concerned about what the is Logger doing.

src/Tracy/Debugger.php
```diff
@@ -400,9 +407,10 @@ public static function setLogger(ILogger $logger)
...
 			self::$logger->directory = & self::$logDirectory; // back compatiblity
 			self::$logger->email = & self::$email;
+			self::$logger->fromEmail = & self::$fromEmail;
```
fprochazka: This right here is a BC break

src/Tracy/Logger.php
```diff
@@ -175,15 +179,16 @@ protected function sendEmail($message)
 	 * @return void
 	 * @internal
 	 */
-	public function defaultMailer($message, $email)
+	public function defaultMailer($message, $email, $fromEmail)
 	{
 		$host = preg_replace('#[^\w.-]+#', '', isset($_SERVER['HTTP_HOST']) ? $_SERVER['HTTP_HOST'] : php_uname('n'));
+		$fromEmail = ($fromEmail) ? $fromEmail : "noreply@$host";
```
fprochazka: why the brackets?

src/Tracy/Debugger.php
```diff
@@ -126,9 +129,10 @@ class Debugger
 	 * @param  mixed   production, development mode, autodetection or IP address(es) whitelist.
 	 * @param  string  error log directory
 	 * @param  string  administrator email; enables email sending in production mode
+	 * @param  string  sender email; in 'From' header when sending email
 	 * @return void
 	 */
-	public static function enable($mode = NULL, $logDirectory = NULL, $email = NULL)
+	public static function enable($mode = NULL, $logDirectory = NULL, $email = NULL, $fromEmail = NULL)
```
fprochazka: I don't like this new argument.

src/Tracy/Debugger.php
Majkl578: But it already is by managing `logDirectory` and `email`.

fprochazka: That's for BC, I don't see any reason to introduce more properties like this on the Debugger.

fifco: But it will not be able to configure it from neon in 'debugger' section, will it?

src/Tracy/Logger.php
fifco: sorry, just a habit, I am goin to correct it

dg:
Logger::$fromEmail is OK, but please do not add new argument to methods.

src/Tracy/Debugger.php
fprochazka: Why not?

fifco: Ok, it does not need to be passed to Logger class but it must stay as property in Debugger class. (I am going to push new commit about this in few sec). Because when I want to configure fromEmail in neon from 'debugger' section it must be declared as static property in Debugger class, or am I wrong?

PR #111, opened 2015-05-07
Sending exception file name in logger mailer

misak113:
It was in earlier version & now it's missing.
Additionally I add optional baseUrl prefix to click in mail to log dir directly.

### DIFF
--- a/src/Tracy/Logger.php
+++ b/src/Tracy/Logger.php
@@ -35,13 +35,17 @@ class Logger implements ILogger
 	/** @var BlueScreen */
 	private $blueScreen;
 
+	/** @var string */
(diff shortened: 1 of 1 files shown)


dg:
Do you store log in public directory??

misak113:
No, it's in private ssh. Example address should be sftp://my.server.cz/my/log/dir/
It could be useful, in our project it is... Second commit is only concept, first commit is pretty critical...

dg:
I don't want to merge this PR because it can cause that users will store logs in public directory. 

You can solve it via own logger `Debugger::getLogger()->mailer = function($message, $emails) { ... }` and get exception file using `$logger->getExceptionFile()`.

PR #115, opened 2015-07-07
BlueScreen: collapse paths usable with files

VasekPurchart:
Right now there is possibility to configure collapsed paths in BlueScreen with `BlueScreen::$collapsePaths`. Since the name of the property contains "paths" I think all paths should be allowed.

But because in the implementation a `/` is added at the end, file paths never match. Adding `/` at the end is fine (actually necessary, because otherwise there will be false positives for dirs - I have added a test for this). With files there is of course similar issue with false positives, so matching must be done "exactly" (not starts with).

So implementations for dirs and files must differ and I've not come up with a different solution (without an API change) other than checking the actual given path if it is a file or directory. I hope that should not be a problem.

### DIFF
--- a/src/Tracy/BlueScreen.php
+++ b/src/Tracy/BlueScreen.php
@@ -177,8 +177,10 @@ class BlueScreen
 	 */
 	public function isCollapsed($file)
 	{
+		$file = strtr($file, '\\', '/');
 		foreach ($this->collapsePaths as $path) {
-			if (strpos(strtr($file, '\\', '/'), strtr("$path/", '\\', '/')) === 0) {
+			$path = strtr($path, '\\', '/');
+			if ($file === $path || strncmp($file, "$path/", strlen($path) + 1) === 0) {
 				return TRUE;
 			}
 		}
--- a/tests/Tracy/Bluescreen.collapsedPaths.phpt
+++ b/tests/Tracy/Bluescreen.collapsedPaths.phpt
@@ -13,6 +13,10 @@ require __DIR__ . '/../bootstrap.php';
 $blueScreen = new Tracy\BlueScreen;
 
 $blueScreen->collapsePaths[] = __DIR__;
+$blueScreen->collapsePaths[] = dirname(__DIR__) . '/bootstrap.php';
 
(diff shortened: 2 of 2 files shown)


JanTvrdik:
This will result in disk IO for every NDB query when ConnectionPanel is used https://github.com/nette/database/blob/187df045a31a22b2d1871e300f819ebada208799/src/Bridges/DatabaseTracy/ConnectionPanel.php#L58

JanTvrdik:
Will this work?

~~~php
$file = strtr($file, '\\', '/');
foreach ($this->collapsePaths as $path) {
	$path = strtr($path, '\\', '/');
	$length = strlen($path);
	if (strncmp($file, $path, $length) === 0 && (!isset($file[$length]) || $file[$length] === '/')) {
		return TRUE;
	}
}
return FALSE;
~~~

VasekPurchart:
> This will result in multiple disk IO operations

That is not necessarily true, because of the [stat cache](http://php.net/manual/en/function.stat.php) and since these are paths from where source codes were probably loaded (if not using opcache), the information might have already been loaded.

> Will this work?

But of course if we can avoid the chance, it's better. I think it will work (I can't come up with a situation where your solution won't work), so I updated the PR, thanks! :+1:

dg:
What about simple if (strncmp("$file/", "$path/", strlen($path) + 1) === 0) {

VasekPurchart:
Yes, that would work too, but I think current solution is more readable, so it is up to you, if you want to do this microoptimalization.

JanTvrdik:
@dg :+1:

dg:
Merged

dg:
(possible BC break: when path in `$collapsePaths` contains leading `/`)

VasekPurchart:
With the submitted 
```php
  if ($file === $path || strncmp($file, "$path/", strlen($path) + 1) === 0) {
```
version, this wouldn't be the issue, would it? Is the "faster" version worth it?

dg:
It would, `"$path/"` will be `/path//`

dg:
Sry, it is not BC break.

VasekPurchart:
Yea, you are right :) Thanks for merging!

PR #116, opened 2015-08-07
severity which throws exception is configurable

ghost:
It is possible configure which severity throw en exception - with respect of the @ operator; some code depends on exceptions from warnings. e.g.
```
try {
    mkdir(...);
} catch(...)
```

### DIFF
--- a/src/Tracy/Debugger.php
+++ b/src/Tracy/Debugger.php
@@ -68,6 +68,12 @@ class Debugger
 	/** @var string|array email(s) to which send error notifications */
 	public static $email;
 
+	/** @var array php severity levels which throw exception */
+	public static $throwOnSeverity = [
+		E_RECOVERABLE_ERROR,
+		E_USER_ERROR,
+	];
+
 	/** {@link Debugger::log()} and {@link Debugger::fireLog()} */
 	const
 		DEBUG = ILogger::DEBUG,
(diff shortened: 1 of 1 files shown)


JanTvrdik:
:-1: If you want to convert errors (e.g. warnings) to exception, you can always set up custom error handler. But changing it globally is usually not good idea as it may break compatibility with some libraries.

Ellesmess:
If I setup custom handler does it suppress Tracy's handler? I want to have Tracy as the only error handler, but a little bit configurable.

Or you mean some sort of "chaining" of error handlers?

JanTvrdik:
> Or you mean some sort of "chaining" of error handlers?

Yes, you can call previous error handler from your error handler, e.g.

~~~php
$prev = set_error_handler(function (...$args) use (& $prev) {
	echo 'my error handler';
	return $prev ? $prev(...$args) : FALSE;
});
~~~

Ellesmess:
Yes, this can be solution, but still don't understand, why this cannot be configurable:
https://github.com/nette/tracy/blob/master/src/Tracy/Debugger.php#L291

I understand, that this
```
if(!error_reporting()) {
	return true;
}
```
can break compatibility (it is probably reason why I'm breaking PHP7 tests). Without this piece is code same, only extended with configuration (and throws exception even @ is used).

dg:
Tracy is debugging tool and it should not change PHP behaviour. Converting E_USER_ERROR and E_RECOVERABLE_ERROR to exceptions is IMHO obsolete and should be removed.

Ellesmess:
Ok, I agree with this argument.

Thanks

PR #122, opened 2015-10-19
Saving and restoring of toggles

dg:


fprochazka:
Is it gonna remember toggles of data dump structures?

dg:
Yes, try it.

fprochazka:
If I have multiple datetime objects in another object, it remembers only the goggle state of the last one and ignores the others.

dg:
Where? In arguments? Debug bar?

fprochazka:
@dg In the panel, in identity dump. I have several datetime properties there.

dg:
@fprochazka can you prepare minimal example?

PR #123, opened 2015-10-29
Css improvement

sallyx:
Added max width for h1


src/Tracy/assets/Bar/bar.css
```diff
@@ -190,6 +190,8 @@ body #tracy-debug {
 	color: #575753;
 	margin: -5px -5px 5px;
 	padding: 0 25px 5px 5px;
+	max-width:700px;
```
dg: missing space after :

PR #125, opened 2015-11-05
Allow stream wrappers

sallyx:
Debugger: allowed using stream wrappers as log directory

This change allows use path for log directory in form  'redis://path/to/cache'


dg:
`is_dir('redis://path/to/cache')` works for these wrappers?

sallyx:
It depends on how the wrapper is written. For example, it works with my [redis wrapper](/sallyx/redis-php-stream-wrapper).

PR #130, opened 2016-01-03
BlueScreen: add renderToFile

Vrtak-CZ:
We are using Tracy with [Monolog](https://github.com/Seldaek/monolog) and we only need "render" bluescreen to file. So I add renderToFile method to BlueScreen (code moved from Logger).

I'm not sure with method name. You maybe have better idea.

### DIFF
--- a/src/Tracy/BlueScreen.php
+++ b/src/Tracy/BlueScreen.php
@@ -68,6 +68,24 @@ class BlueScreen
 
 
 	/**
+	 * @param \Exception|\Throwable
+	 * @param string file path
+	 * @return void
+	 */
+	public function renderToFile($exception, $file)
+	{
+		if ($handle = @fopen($file, 'x')) { // @ file may already exist
+			ob_start(); // double buffer prevents sending HTTP headers in some PHP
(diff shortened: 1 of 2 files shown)


fprochazka:
Nice :+1:

src/Tracy/BlueScreen.php
```diff
@@ -68,6 +68,24 @@ public function render($exception)
 
 
 	/**
+	 * @param \Exception|\Throwable
+	 * @param string file path
+	 * @return void
+	 */
+	public function renderToFile($exception, $file)
+	{
+		if ($handle = @fopen($file, 'x')) { // @ file may already exist
```
dg: IMHO this conditions should stay in Logger.

Vrtak-CZ: Ok but if I move that to Logger I must pass resource to this method from Logger or we should have "copy&paste" code in Logger and in this method. But file as resource for this method doesn't looks good for me (it's just feeling).

Do you have better idea?

dg: File should be opened twice, via `@fopen($file, 'x')` in logger and `fopen($file, 'w')` in Bluescreen.

PR #83, opened 2014-11-22
Tracy\Bar's output can be disabled (closes #82)

hranicka:
Closes #82

PR #132, opened 2016-01-10
Added disable() method to Tracy/Bar to disable the debug bar

wouterds:


JanTvrdik:
ref https://github.com/nette/tracy/pull/83

wouterds:
👍

PR #135, opened 2016-02-11
Dumper: use __debugInfo if present by default to export object

vojtech-dobes:
This is resurrection of https://github.com/nette/nette/pull/1415.

Tests are added.

I've also added possibility to switch this off with `Dumper::DEBUGINFO => FALSE`. I've chosen `TRUE` as default because it feels to me not only as new feature, but rather matching default behavior of PHP. <del>Last point: I've enabled it only for PHP >=5.6 where this magic method is actually supported by PHP.</del>

### DIFF
--- a/src/Tracy/Dumper.php
+++ b/src/Tracy/Dumper.php
@@ -22,7 +22,8 @@ class Dumper
 		COLLAPSE_COUNT = 'collapsecount', // how big array/object are collapsed? (defaults to 7)
 		LOCATION = 'location', // show location string? (defaults to 0)
 		OBJECT_EXPORTERS = 'exporters', // custom exporters for objects (defaults to Dumper::$objectexporters)
-		LIVE = 'live'; // will be rendered using JavaScript
+		LIVE = 'live', // will be rendered using JavaScript
+		DEBUGINFO = 'debuginfo'; // use magic method __debugInfo if exists (defaults to true)
 
 	const
 		LOCATION_SOURCE = 0b0001, // shows where dump was called
@@ -94,6 +95,7 @@ class Dumper
 			self::COLLAPSE => 14,
 			self::COLLAPSE_COUNT => 7,
 			self::OBJECT_EXPORTERS => NULL,
+			self::DEBUGINFO => TRUE,
 		];
 		$loc = & $options[self::LOCATION];
 		$loc = $loc === TRUE ? ~0 : (int) $loc;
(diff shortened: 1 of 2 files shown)


hrach:
The current master is PHP 5.6+ only, so I think the condition may be removed :)

vojtech-dobes:
> The current master is PHP 5.6+ only, so I think the condition may be removed :)

Ah, indeed, good point :). I will gladly do that.

dg:
Uses __debugInfo anyone?

vojtech-dobes:
> Uses __debugInfo anyone?

We do, can't speak for anyone else :).

staabm:
For aerys we also use it.

enumag:
I certainly will when this is merged.

JanTvrdik:
:-1: On using debug info by default. The only real experience with debug info I have is that it used to cause a lot of [strange failures](https://github.com/nextras/orm/commit/bf92fd1735c0390e9a9bb9bc3263871fc917e767), not sure whether this has been fixed in PHP 7.

hrach:
Personally, I'd love to see the support for this. The mentioned php bug is a little problem here, so I can imagine this as a non-default bahavior. Would be great to be able enable/disable this globally.

fprochazka:
I think making objects dump their guts is violation of SRP, in the same way adding a public getter to be able to test a state of object is. 

Imho the right way to solve this is this https://github.com/nette/nette/pull/1090#issuecomment-53957680

vojtech-dobes:
What I would like to use this for can be seen in attached test case. I don't want to expose internals of object, rather conceal them from being dumped somewhere in logs... So I will definitely use `__debugInfo`, but it would be nice if Tracy would understand it too and I wouldn't have to write specific code for it.

hrach:
@fprochazka This is internal functionality, the purpose is for debugging. Such method has nothing to do with SRP.

fprochazka:
@hrach I disagree. But then again, I'm not strongly against this feature. Just wanned to give my two cents and point to an existing and IMHO better way of dumping&debugging object data.

vojtech-dobes:
@fprochazka It can, *additionally* to `__debugInfo` :).

dg:
I am really not sure whether it is useful to support __debugInfo, but in the meantime you can emulate it:

```php
Dumper::$objectExporters[''] = function($obj) {
	return method_exists($obj, '__debugInfo') ? $obj->__debugInfo() : (array) $obj;
};
```

vojtech-dobes:
@dg That's nice - may I ask what's the trick? Why does `NULL` match every object?

JanTvrdik:
@vojtech-dobes Have you seen https://github.com/nette/tracy/commit/803753373118f4485b139a065397a190df9aab27?

vojtech-dobes:
> @vojtech-dobes Have you seen 8037533?

Thanks. I've discovered it today when exploring new release - clear now.

PR #137, opened 2016-02-16
Pass array for $args to formatHtml

vokiel:
Method `self::formatHtml` from `Tracy\Helpers` was invoked incorrectly.
This was causing PHP Warning: `Variable passed to each() is not an array or object` and due to this the errors list was incomplete:

Before:
tracy_errors

After fix:
tracy_errors_fixed


vokiel:
My bad, was running it on PHP 5.5

PR #141, opened 2016-02-23
Dumper: fix possible E_NOTICE

danielkurecka:


Majkl578:
Empty property name is invalid state, it results in notices or even fatal error in many cases.

PR #133, opened 2016-03-02
Feature Request: Add last error panel to bluescreen

JanTvrdik:
In PHP it is quite common to use "shut-up + escalating error to exception" pattern ([example](https://github.com/nette/utils/blob/9cd7d72e92d613407bf8d336a49225566c748a79/src/Utils/FileSystem.php#L24-L29)). Unfortunately the original error message is lost unless `error_get_last()['message']` is manually appended to exception message.

What about adding a simple panel to bluescreen to show the last error? The error may not always be related. We may attempt to do some magic to filter the relevant error but I'm not sure it's worth it.

---

Related: https://github.com/nette/utils/pull/94

### DIFF
--- a/src/Tracy/BlueScreen.php
+++ b/src/Tracy/BlueScreen.php
@@ -95,6 +95,8 @@ class BlueScreen
 		$skipError = $sourceIsUrl && $exception instanceof \ErrorException && !empty($exception->skippable)
 			? $source . (strpos($source, '?') ? '&' : '?') . '_tracy_skip_error'
 			: NULL;
+		$lastError = (!$exception instanceof \ErrorException) ? error_get_last() : NULL;
+
 		require $template;
 	}
 
--- a/src/Tracy/assets/BlueScreen/content.phtml
+++ b/src/Tracy/assets/BlueScreen/content.phtml
@@ -174,6 +174,23 @@ $code = $exception->getCode() ? ' #' . $exception->getCode() : '';
 		<?php endif ?>
 
 
+		<?php if ($lastError): ?>
+		<div class="panel">
+		<h2><a data-tracy-ref="^+" class="tracy-toggle tracy-collapsed">Last error</a></h2>
+		<div class="tracy-collapsed inner">
+
+		<h3><?= Helpers::errorTypeToString($lastError['type']) ?>: <?= htmlspecialchars($lastError['message'], ENT_IGNORE, 'UTF-8') ?></h3>
+		<?php if (isset($lastError['file']) && is_file($lastError['file'])): ?>
+			<p><?= Helpers::editorLink($lastError['file'], $lastError['line']) ?> <a data-tracy-ref="^+" class="tracy-toggle">source</a>&nbsp;</p>
+			<div><?= self::highlightFile($lastError['file'], $lastError['line']) ?></div>
+		<?php else: ?>
+			<p><i>inner-code</i><?php if (isset($lastError['line'])) echo ':', $lastError['line'] ?></p>
+		<?php endif ?>
+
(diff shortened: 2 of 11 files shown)


JanTvrdik:
Any thoughts?

fprochazka:
@JanTvrdik maybe it would be nice to hide the error, if the rendered exception is ErrorException created by nette handler and the error would be exactly the same?

JanTvrdik:
@fprochazka I know but I don't see a simple and reliable way to write the condition. You may check ErrorException + file + line but it may have some false positives.

JanTvrdik:
ping @dg, is this ok? should i fix tests?

src/Tracy/assets/BlueScreen/bluescreen.phtml
```diff
@@ -195,6 +195,23 @@ $code = $exception->getCode() ? ' #' . $exception->getCode() : '';
 		<?php endif ?>
 
 
+		<?php if ($error = error_get_last()): ?>
+		<div class="panel">
+		<h2><a data-tracy-ref="^+" class="tracy-toggle tracy-collapsed">Last error</a></h2>
+		<div class="tracy-collapsed inner">
+
+		<h2><?= Helpers::errorTypeToString($error['type']) ?>: <?= htmlspecialchars($error['message'], ENT_IGNORE, 'UTF-8') ?></h2>
```
dg: Maybe H3?

src/Tracy/assets/BlueScreen/bluescreen.phtml
```diff
@@ -195,6 +195,23 @@ $code = $exception->getCode() ? ' #' . $exception->getCode() : '';
 		<?php endif ?>
 
 
+		<?php if ($error = error_get_last()): ?>
```
dg: This should be saved sooner, because Tracy can rewrite it.

src/Tracy/assets/BlueScreen/bluescreen.phtml
```diff
@@ -195,6 +195,23 @@ $code = $exception->getCode() ? ' #' . $exception->getCode() : '';
...
+		<div class="panel">
+		<h2><a data-tracy-ref="^+" class="tracy-toggle tracy-collapsed">Last error</a></h2>
+		<div class="tracy-collapsed inner">
+
+		<h2><?= Helpers::errorTypeToString($error['type']) ?>: <?= htmlspecialchars($error['message'], ENT_IGNORE, 'UTF-8') ?></h2>
+		<?php if (isset($error['file']) && is_file($error['file'])): ?>
+			<p><?= Helpers::editorLink($error['file'], $error['line']) ?> <a data-tracy-ref="^+" class="tracy-toggle tracy-collapsed">source</a>&nbsp;</p>
```
dg: This panel is by collapsed, so the source code can be visible immediately.

JanTvrdik:
:tired_face: I've put hours of energy into an attempt to fix the tests and it's still not done. This is ridiculous. :tired_face:

src/Tracy/BlueScreen.php
```diff
@@ -95,6 +95,8 @@ private function renderTemplate($exception, $template)
 		$skipError = $sourceIsUrl && $exception instanceof \ErrorException && !empty($exception->skippable)
 			? $source . (strpos($source, '?') ? '&' : '?') . '_tracy_skip_error'
 			: NULL;
+		$lastError = (!$exception instanceof \ErrorException) ? error_get_last() : NULL;
+
```
dg: `$lastError = $exception instanceof \ErrorException ? NULL : error_get_last();` seems better

PR #147, opened 2016-03-07
dump php://input in bluescreen

mzk:
add support for PATCH requests

- [ ] fix tests

Is this solution ok?

### DIFF
--- a/src/Tracy/BlueScreen.php
+++ b/src/Tracy/BlueScreen.php
@@ -63,6 +63,8 @@ class BlueScreen
 			? $source . (strpos($source, '?') ? '&' : '?') . '_tracy_skip_error'
 			: NULL;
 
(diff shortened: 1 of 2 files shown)


JanTvrdik:
:-1:

milo:
The body can be in MB. Some insight would be handy, but this is too much.

JanTvrdik:
I changed my mind, it would be useful. How about some when request content-type is JSON show first 1kB?

mzk:
I solved in BasePresenter.. 
//example
```php
try {
	$_SERVER['API_REQUEST'] = Json::decode($this->getHttpRequest()->getRawBody(), Json::FORCE_ARRAY); //for Tracy log
} catch (JsonException $e) {
	//solve in another place
}
```

`$_SERVER` is dumped in tracy by default.

PR #148, opened 2016-03-07
Undefined index of source

OzzyCzech:
We found many notices in our log...

PHP Notice: Undefined offset: 941 in /var/www/testomato/htdocs/vendor/tracy/tracy/src/Tracy/BlueScreen.php:140


matej21:
The origin of this error is probably elsewhere. I've experienced similar issue and I found this: https://github.com/nette/latte/commit/49c527c617d2ab97b3b7b7c3f736ed5cc511fa37 (line number was overwritten)

PR #155, opened 2016-03-26
evalScript: only evaluate *javascript* scripts

jahudka:
the script tag can be used for other things than javascript - script tags with a different "type" than text/javascript shouldn't be assumed to be javascript

### DIFF
--- a/src/Tracy/assets/Bar/bar.js
+++ b/src/Tracy/assets/Bar/bar.js
@@ -401,10 +401,12 @@
 
 	function evalScripts(elem) {
 		[].forEach.call(elem.querySelectorAll('script'), function(script) {
-			(window.execScript || function(data) {
-				window['eval'].call(window, data);
-			})(script.innerHTML);
-			script.parentNode.removeChild(script);
+			if (!script.hasAttribute('type') || script.type === 'text/javascript' || script.type === 'application/javascript') {
+				(window.execScript || function (data) {
+					window['eval'].call(window, data);
+				})(script.innerHTML);
+				script.parentNode.removeChild(script);
+			}
 		});
 	};
 


Majkl578:
text/javascript is obsoslete, it should be application/javascript.

jahudka:
@Majkl578 well to really give as much compatibility as we can we should really support _both_.. it's just that text/javascript works everywhere, whereas application/javascript doesn't (at least IE <= 8), but that doesn't necessarily matter in this context (Tracy doesn't support IE <= 8 I think).

milo:
@jahudka Could you change commit message to something like: `bar.js: eval only javascript scripts`? Then ready to merge.

dg:
ie what about  `[].forEach.call(elem.querySelectorAll('script[data-value="text/javascript"], script:not([type])'), function(script) {`

jahudka:
@dg that wouldn't support `application/javascript`, it'd have to be `[].forEach.call(elem.querySelectorAll('script[type="text/javascript"], script[type="application/javascript"], script:not([type])'), function(script) {`; moreover, there's the question of querySelectorAll support and performance - getElementsByTagName is going to be a lot faster and it's supported everywhere.

dg:
I'll change it to getElementsByTagName

dg:
Screw it, it is not faster http://jsperf.com/getelementsbytagname-a-0-vs-queryselector-a/25

jahudka:
Not really, your test is comparing something else, this is the correct test:
http://jsperf.com/queryselectorall-vs-getelementsbytagname-performance

dg:
Thats right

jahudka:
getElementsByTagName is actually *two orders of magnitude* faster than querySelectorAll :)

dg:
No, you should measure iteration too. Try this http://jsperf.com/getelementsbytagname-a-0-vs-queryselector-a/13

dg:
(Btw https://github.com/nette/forms/blob/master/src/assets/netteForms.js#L568 is intention, *change* is not bubbling in IE < 9).

jahudka:
Well, even if you include iteration in the test, the result is still an order of magnitude in favour of getElementsByTagName..

Re change not bubbling in IE < 9, well, it won't break anything, will it? And some JS frameworks emulate / polyfill various non-standard behaviours in older browsers, including which events do bubble and which don't; I'm not sure if these emulations would trigger the native event, but I'd err on the side of caution here rather than write code which could possibly break..

PR #160, opened 2016-05-05
Delay panel rendering

JanTvrdik:
Similarly to how we delay rendering of invisible dumps we should delay rendering of invisible panels and render them the first time they became visible. This is especially important when bar panel contains a lot of HTML (for example query log).

What do you think?

### DIFF
--- a/src/Tracy/Debugger.php
+++ b/src/Tracy/Debugger.php
@@ -550,7 +550,7 @@ class Debugger
 		if (!self::$productionMode) {
 			static $panel;
 			if (!$panel) {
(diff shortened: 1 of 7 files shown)


JanTvrdik:
@dg This is now ready to merge from my POV.

JanTvrdik:
:tired_face: It breaks toggle persisting in window-mode, I'll look into it.

dg:
Great!

adrianbj:
Thanks @JanTvrdik - this is awesome!

dg:
@JanTvrdik there is problem when panel contains dumps with LIVE => TRUE, it is not inited.

PR #161, opened 2016-05-05
Tracy does not preserve panel id when moving panel to new window

JanTvrdik:
Current behavior breaks for example [Nextras Mail Panel](https://github.com/nextras/mail-panel/blob/e238dd9d0311f941f3f09433094059b15c49d4f4/src/MailPanel.latte#L1-L40).

### DIFF
--- a/src/Tracy/assets/Bar/bar.js
+++ b/src/Tracy/assets/Bar/bar.js
@@ -147,7 +147,7 @@
 			+ '<script src="' + escape(document.getElementById('tracy-debug-script').src) + '" onload="Tracy.Dumper.init()" async><\/script>'
 			+ '<body id="tracy-debug">'
 		);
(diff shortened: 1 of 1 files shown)


JanTvrdik:
Just found out that the panel id was preserved in Tracy 2.3 and this behavior was removed in https://github.com/nette/tracy/commit/13f854d4c11221add4697121e183a147732bb13b. Do you remember why was it required, @dg?

dg:
Yes I do. I could not figure out what purpose of ID was, so I put it away…

JanTvrdik:
I updated the nextras mail panel to [not rely on the id](https://github.com/nextras/mail-panel/commit/fa74013cce43a2520810230a698e90c1eeba7e67) but dropping the ID still feel like a bug.

PR #162, opened 2016-05-06
Fixed panel error handling

JanTvrdik:


src/Tracy/Bar.php
```diff
@@ -80,6 +80,14 @@ public function render()
 		$obLevel = ob_get_level();
 		$panels = [];
 
+		set_error_handler(function ($severity, $message, $file, $line, $context) {
+			if (error_reporting() & $severity) {
+				$e = new \ErrorException($message, 0, $severity, $file, $line);
+				$e->context = $context;
```
dg: Is $context somehow used?

PR #163, opened 2016-05-06
bar.css: padding of panel in window-mode is consistent with other modes

JanTvrdik:
*While working on certain tracy panel…*

I discovered annoying inconsistency in how panels are rendered in window mode.

### DIFF
--- a/src/Tracy/assets/Bar/bar.css
+++ b/src/Tracy/assets/Bar/bar.css
@@ -9,7 +9,7 @@
 }
 
 body#tracy-debug {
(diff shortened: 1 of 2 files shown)


src/Tracy/assets/Bar/bar.css
```diff
@@ -9,7 +9,7 @@
 }
 
 body#tracy-debug {
-	margin: 5px 5px 0;
+	margin: 0;
```
JanTvrdik: I hope this does not break anything.

PR #165, opened 2016-05-06
bar.js: fixed evaluating script tags in panels in window-mode

JanTvrdik:
*Still working on certain tracy panel…*

This changes evalScripts to NOT remove the evaluated scripts (which was introduced in https://github.com/nette/tracy/commit/63b03725b1108c8627567512d875c349c2b8689b). This likely breaks some other behavior.

### DIFF
--- a/src/Tracy/assets/Bar/bar.js
+++ b/src/Tracy/assets/Bar/bar.js
@@ -148,6 +148,7 @@
 			+ '<body id="tracy-debug">'
 		);
 		doc.body.innerHTML = '<div class="tracy-panel tracy-mode-window">' + this.elem.innerHTML + '<\/div>';
+		evalScripts(doc.body, win);
 		if (this.elem.querySelector('h1')) {
 			doc.title = this.elem.querySelector('h1').innerHTML;
 		}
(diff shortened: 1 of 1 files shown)


JanTvrdik:
@dg This should be fixed before 2.4 is released as it is directly affected by changed in introduces in 2.4 and therefore may require refactoring some of Tracy ajax-related logic.

dg:
Why is needed to not remove evaluated scripts?

JanTvrdik:
Because if I remove them I can't evaluate them again in the scope of the new window. You can easily test the problem with Nextras Mail Panel.

dg:
Evaled scripts should be somehow marked as evaled (class, dataattribute) and ignored next time.

JanTvrdik:
Thanks, I tested Nextras Mail Panel and it works perfectly. Should I send another PR for Tracy 2.3 or do you want to backport it yourself?

dg:
I'll try to backport it.

dg:
Ha, it is not so easy… I you really want to backport it, do it please :-)

PR #166, opened 2016-05-08
Logger: can now report errors by other means than email

kravco:
My aim is to be able to report errors via text messages. But idea is to allow any other means for reporting, e.g. create issue in internal reporting system--the advantage would be, that it would contain all the necessary information for future reference.

Possibly in the future, email would lose it's special status because--in the end--it is just another reporting channel, no different from others, mentioned earlier.

If agreed, I suggest traditional slow transition, with:
  1. full backwards compatibility
  2. bc with triggering deprecated warning
  3. release with only new functionality

I'm ready to prepare these steps too.

### DIFF
--- a/src/Tracy/Logger.php
+++ b/src/Tracy/Logger.php
@@ -28,6 +28,9 @@ class Logger implements ILogger
 	/** @var callable handler for sending emails */
 	public $mailer;
 
+	/** @var array list of additional channels for reporting errors */
+	private $reportChannels = [];
+
 	/** @var BlueScreen */
 	private $blueScreen;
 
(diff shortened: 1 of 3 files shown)


JanTvrdik:
:-1: I'd suggest using Monolog instead.

dg:
Welcome back @kravco :-)

I agree that reporting via email is too simple. On the other hand, I do not know if it makes sense to develop logging part of Tracy, because it is not primarily a logging tool and there are better tools. Usually, when someone wants to use more sophisticated logging, he creates its own Logger, while the reportChannels are not enough complex solution.

kravco:
Yeah, it's been a long time :-)

Ok, I understand. I can write my own logger for this.

PR #168, opened 2016-05-11
Added support of generating unique html trace of every error without …

cniry:
…linking to old logs. (Logger::$allowLinkOldTrace)


dg:
This is imho very egde use case, so better way is to create own Logger that inherits from default Logger.

PR #164, opened 2016-05-13
Fix opacity hover effect on panels

JanTvrdik:
All browsers except for Chrome have trouble rendering them.

### DIFF
--- a/src/Tracy/assets/Bar/bar.css
+++ b/src/Tracy/assets/Bar/bar.css
@@ -110,7 +110,7 @@ body #tracy-debug {
 	z-index: 30000;
 	opacity: .9;
 	transition: opacity 0.2s;
-	will-change: opacity;
+	will-change: opacity, right, bottom;
 
 	border-radius: 3px;
 	box-shadow: 1px 1px 10px rgba(0, 0, 0, .15);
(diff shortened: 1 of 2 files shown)


hrach:
I have no idea which issue are you trying to solve :-) For me, the opacity works ok in Firefox :-) I'm just curious, could you make a little screencast? :)

JanTvrdik:
@hrach  [recording-2016-05-13-13310.zip](https://github.com/nette/tracy/files/263067/recording-2016-05-13-13310.zip)

dg:
On Windows it is fine. Btw I had similar problem on one website with Chrome and solution was `	-webkit-backface-visibility: hidden`.

JanTvrdik:
@dg I have Windows 10.

dg:
Hmmm, me too. Windows 10, Firefox 45.0.2, now updated to 46.0.1 and no problems.

JanTvrdik:
Latest Firefox, clean profile without any option changes, the transition without will-change is broken both on panels and bar.

tracy-ff

dg:
What about change max opacity to 0.99?

JanTvrdik:
What's the problem with will-change? Changing max opacity to 0.99 does not help.

dg:
will-change. is ok, just this PR seems like bigger fix.

JanTvrdik:
@dg Should I move the dragging optimization to another PR? I'm working now on bugfixing + optimizing `moveConstrains`

hrach:
oh yeah, this will be probably some bug of graphic drivers & firefox, still :+1: if it works for you.

JanTvrdik:
@dg This is now ready to merge from my point of view.

The last commit introduces a behavior change – panel can no longer be partially (20%) out of the window – this fixes issue with panels that were impossible to move because drag handler (h1) was out of the window.

dg:
Thanks!

PR #170, opened 2016-05-15
Add $editorMapping support

adrianbj:
As per: https://forum.nette.org/en/24404-editor-link-in-virtualbox

and 

https://forum.nette.org/en/25719-public-access-to-modify-returned-file-from-findlocation

There are multiple scenarios when this is useful. Thanks for considering.

### DIFF
--- a/src/Tracy/Debugger.php
+++ b/src/Tracy/Debugger.php
@@ -91,6 +91,9 @@ class Debugger
 	/** @var string URI pattern mask to open editor */
 	public static $editor = 'editor://open/?file=%file&line=%line';
 
+	/** @var array replacements in path */
+	public static $editorMapping = array();
+
 	/** @var string command to open browser (use 'start ""' in Windows) */
 	public static $browser;
 
--- a/src/Tracy/Helpers.php
+++ b/src/Tracy/Helpers.php
@@ -20,6 +20,9 @@ class Helpers
 	 */
 	public static function editorLink($file, $line = NULL)
 	{
+
+		$file = strtr($file, Debugger::$editorMapping);
+
 		if ($editor = self::editorUri($file, $line)) {
 			$file = strtr($file, '\\', '/');
 			if (preg_match('#(^[a-z]:)?/.{1,50}$#i', $file, $m) && strlen($file) > strlen($m[0])) {
@@ -45,6 +48,9 @@ class Helpers
 	 */
 	public static function editorUri($file, $line = NULL)
 	{
+
+		$file = strtr($file, Debugger::$editorMapping);
+
 		if (Debugger::$editor && $file && is_file($file)) {
 			return strtr(Debugger::$editor, ['%file' => rawurlencode($file), '%line' => $line ? (int) $line : 1]);
 		}


rostenkowski:
This feature would be nice for anybody using some remote/virtual/production environment. 

If I am reading that build information right there is only used the `array(...)` syntax instead of `[...]`.

dg:
It is already implemented here https://github.com/nette/tracy/blob/master/tools/open-in-editor/open-editor.js#L25-L27

adrianbj:
@dg - I am sorry if I am missing something, but doesn't that open-editor.js file only get used on Windows systems?

adrianbj:
As a follow up, the other problem with the open-editor.js approach is that it I assume it affects all links to editor:// not just those coming from Tracy. This may not always be desired.

@dg - Just wondering if you completely opposed to implementing something along the lines of this PR? I would just like to have an idea as to whether I need to continue to hack the Tracy core to make it functional for my needs. Perhaps you would consider some other approach that would allow editor filepath mapping that works on all systems, is just for Tracy links, and is easily configurable by the user or the dev packaging Tracy for specific purposes

Thanks for reconsidering.

dg:
You are right that open-editor.js is platform specific, I'll add it.

adrianbj:
Thanks @dg - the problem is that it doesn't work for me. The substr / strlen combination only works if the replacement is right at the beginning of the path.

Is there any reason you didn't like the `strtr` approach which allows for much more flexibility and no need for loops? I actually thought that strtr was generally faster anyway.

adrianbj:
Oh and one more comment - your approach only changes the link, but not the displayed URL. Perhaps that is desirable in some situations, but in my case I'd also prefer it to be changed as well, which is why I also has the strtr line in the editorLink method.

Perhaps this could be a configurable option.

dg:
I supposed that you want to change begin of path - or not? Can you describe you use case?

adrianbj:
I guess that might be the case for some users. In my situation I need to convert from a path where there are compiled files to the path of the original files. Files are compiled at runtime if there have been any changes made.

For example, I need to convert:
```
/Users/ajones/Sites/pwstable/site/assets/cache/FileCompiler/site/templates/home.php
```
to
```
/Users/ajones/Sites/pwstable/site/templates/home.php
```

So my $editorMapping looks like this:
```
Debugger::$editorMapping = array('/site/assets/cache/FileCompiler' => '');
```

which works perfectly with my original strtr approach.

adrianbj:
Just to clarify - I guess I could get the full path and replace that with the fullpath without the `/site/assets/cache/FileCompiler` but my approach still seems more flexible unless you have a good reason for not wanting to allow that.

I would also like your thoughts on the ability to modify the displayed URL as well as the actual link. I was thinking maybe a new option:
```
$editorMappingReplaceDisplayedUri = true;
```
It's a very long and ugly variable, but maybe that's ok?

dg:
Ok, repushed.

adrianbj:
Thanks - that works much better I think.

Quick question - I see you moved the strtr in the editorLink() method so that it's inside the "if". Does it make sense that we wouldn't also want to translate the $file in the "else"
```
return self::formatHtml('<span>%</span>', $file . ($line ? ":$line" : ''));
```

I don't really have any experience calling methods from the OutputDebugger class (which I think is the only situation when that conditional would default to "else"), so maybe it wouldn't be appropriate, but thought I should ask.

Sorry for taking up so much of your time on this PR.

dg:
It is inside because 1) `is_file($file)` must be called with original file name and 2) we don't want to call strtr two times

adrianbj:
Right - makes sense to prevent the calling of strtr twice.

I still don't know about:
```
return self::formatHtml('<span>%</span>', $file . ($line ? ":$line" : ''));
```

If this is triggered it will return the $file with the original path. I haven't come across this scenario, but I'll keep an eye out for it.

dg:
Oh, now I see! Fixed :-)

adrianbj:
Perfect, I think that should cover everything now - thank you!

PR #174, opened 2016-05-17
bar.js: fixed evaluating script tags in panels in window-mode

JanTvrdik:
I copy pasted the code to avoid refactoring.

---

The broken tests are not caused by this PR.


dg:
simple :-)

JanTvrdik:
Would you prefer smart?

PR #181, opened 2016-05-20
AJAX Tracy does work over HTTPS

JanTvrdik:
I don't know why. I'll look into it.


JanTvrdik:
It was a [Caddy](https://caddyserver.com/) thing. For some reason it returns lower-cased HTTP headers when request is sent over HTTPS.

dg:
:-)

Issue #184, opened 2016-05-20
Panel height needs to not exceed the height of the viewport


adrianbj:
This can easily become a problem when you have the console open and a panel has a lot of height. The panel needs to be resized to fit, which then triggers scrolling on tracy-inner.

At the moment it's impossible to see all the content as the panel just overflows off the viewport.

JanTvrdik:
That's seems more like a problem with specific panel. Can you post more specific example?

adrianbj:
I don't think it is panel specific - you can replicate with the System Info panel
![screen shot 2016-05-20 at 7 21 42 am](https://cloud.githubusercontent.com/assets/871211/15430773/9eb7f52a-1e5b-11e6-9b4f-434623008293.png)

Obviously I have set an extremely small viewport height for this, but for panels with more content this happens easily with the browser dev console set to a normal height.

I have quickly hacked around this by setting the height for certain custom panels and a percentage max-height which triggers the scrolling of tracy-inner. This kinda works, but it really needs a core JS solution so that the height of the panel can fit the content if there is enough viewport height.

I might find some time to work on a proper solution at some point but I haven't had a good look through  bar.js yet.

JanTvrdik:
This sort-of fixes the problem, but it is not nice and there are likely edge cases it does not handle.

~~~js
window.addEventListener('resize', function () {
	forEach(document.querySelectorAll('#tracy-debug .tracy-mode-float'), function (panel) {
		var inner = panel.querySelector('.tracy-inner');
		var padding = panel.offsetHeight - inner.offsetHeight;
		inner.style.maxHeight = Math.min(document.documentElement.clientHeight - padding, 500) + 'px';
	});
});
~~~

adrianbj:
Thanks - a couple of thoughts. 

1. I think a little bit of a margin top and bottom of the panel would be nice.
2. It also needs to be triggered on document ready, not just resize, because when developing you often have the console open while reloading a page. Currently you need to trigger a window resize to get the panels to fit.

dg:
Possible solution https://github.com/nette/tracy/pull/188

adrianbj:
Thanks - this seems to work well. My only suggestion is to prevent the bottom of the panels from being covered by the debug bar.

adrianbj:
[...]

PR #188, opened 2016-05-21
panels for too small viewport

dg:


JanTvrdik:
Super cool but unfortunately a BC breaks for every panel that assumed scrollbars on `.tracy-inner`. So for example Nextras Mail Panel is broken by this change.

adrianbj:
@JanTvrdik  - oh yeah - I didn't see that - all my custom panels use tracy-inner like that too. Now the title at the top of the panel scrolls with the content. I actually tried this approach initially when trying to deal with this and came to the same realization :)

dg:
What exactly is the problem?

adrianbj:
There may be more to it in @JanTvrdik's case, but for me it is simply that you can't keep the panel <h1> header sticky at the top because the entire panel is now scrolling, not just tracy-inner. This prevent easy access to by able to move, close, and swap to a popup window without having to scroll to the top of the panel.

JanTvrdik:
@dg I need to have (for Tracy 2.3) reserved space for vertical scrollbar to prevent the ugly horizontal scrollbar.

~~~css
#tracy-debug .tracy-panel.tracy-mode-peek .tracy-inner.nextras-mail-panel,
#tracy-debug .tracy-panel.tracy-mode-float .tracy-inner.nextras-mail-panel {
	overflow-y: scroll;  /* reserve space for scrollbar */
}
~~~

This will make the panel in Tracy 2.4 visually inconsistent with others when viewport is large (because it will keep scrolling on .tracy-inner) and broken with ugly triple scrollbar when viewport is small

image

dg:
@adrianbj  It is intention because in small viewport header takes big space for itself. Dragging is now possible via border.

JanTvrdik:
> It is intention because in small viewport header takes big space for itself.

The problem is that it makes the header hard to access even when the viewport is large (height 500px+) and the panel scrolls simple because there are lot of data.

dg:
Its pity. I don't know how to solve  #184.

adrianbj:
I am also having the horizontal scroll issue and also agree that you need access to the header all the time.

I think that https://github.com/nette/tracy/issues/184#issuecomment-220625501 is a decent start to the solution - it just also needs to be called when the panel is displayed, not just when the browser is resized.

dg:
Problem it that you cannot define max-width: 500px and max-width: 100hv at the same time.

JanTvrdik:
@dg Just an idea – what about applying `calc(100vh - 22px)` based on media query on height?

dg:
@JanTvrdik YES!

JanTvrdik:
:+1: Work quite well. Now I'm just not sure whether we need the first commit

adrianbj:
I am finding that I need 75, rather than 25:
```
max-height: calc(100vh - 75px);
```
It is still hidden behind the debug bar, but at least if you move left of the bar, you can see it all. With 25 I still have 50x below the bottom of the viewport.

Chrome/OSX

JanTvrdik:
@adrianbj The updated commit uses `55px`. It assumes that outside of `.tracy-inner` is only a single `<h1>` element.

adrianbj:
With `55px` I still can't see the bottom of the panel.

screen shot 2016-05-21 at 12 37 14 pm

That's the unmodified System Info which only has a single `<h1>`

I just checked on Firefox and it actually looks ok.

The other problem highlighted by that screenshot is that when vertical scrollbar appears, the horizontal one also kicks in.

JanTvrdik:
> I still can't see the bottom of the panel.

That's because of the horizontal scrollbar. The only reliable solution AFAIK is to add `overflow-y: scroll` to `.tracy-inner`.

adrianbj:
Actually it's not just because of the horizontal scrollbar - here it is in firefox - scrollbar is there, but you can still see the bottom of the panel - note the rounded corners that can't be seen in the Chrome screenshot.

screen shot 2016-05-21 at 12 56 52 pm

dg:
@adrianbj can you try to set overflow-y: scroll to .tracy-inner ?

adrianbj:
Obviously that gets rid of the horizontal scrollbar, but it still doesn't let me see the bottom of the panel in Chrome.

dg:
55px works fine in my Chrome 51 on Windows, your Firefox on Mac, but it is different on your Chrome/Mac, right? And headers are exatly the same in both pictures. Weird…

PR #171, opened 2016-05-21
tracy.logger definition can be changed in config

JanTvrdik:
https://forum.nette.org/cs/26317-jak-nastavit-vlastni-logger#p174402

### DIFF
--- a/src/Bridges/Nette/TracyExtension.php
+++ b/src/Bridges/Nette/TracyExtension.php
@@ -76,6 +76,11 @@ class TracyExtension extends Nette\DI\CompilerExtension
 			}
 		}
 
+		$logger = $builder->getDefinition($this->prefix('logger'));
+		if ($logger->getFactory()->getEntity() !== 'Tracy\Debugger::getLogger') {
+			$initialize->addBody($builder->formatPhp('Tracy\Debugger::setLogger(?);', [$logger]));
+		}
+
 		if ($this->debugMode) {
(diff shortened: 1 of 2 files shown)


src/Bridges/Nette/TracyExtension.php
```diff
@@ -76,6 +76,11 @@ public function afterCompile(Nette\PhpGenerator\ClassType $class)
 			}
 		}
 
+		$logger = $builder->getDefinition($this->prefix('logger'));
+		if ($logger->getFactory()->getEntity() !== 'Tracy\Debugger::getLogger') {
+			$initialize->addBody($builder->formatPhp('Tracy\Debugger::setLogger(?);', ['@' . $this->prefix('logger')]));
```
dg: this is possible too: `[$this->prefix('@logger')]));`

JanTvrdik: cool, I did not know that

kukulich:
@JanTvrdik Nemělo by to počítat i dalšími formáty `$entity`, které podporuje `Nette\DI\Statement`?

dg:
In aftercompile() are entites normalized (in 2.4)

JanTvrdik:
@kukulich Could you elaborate? 

---

@dg Looking at the code once again, is possible that `$logger->getFactory()` will return `NULL` and `getEntity()` call will fail?

dg:
I think not

PR #190, opened 2016-05-22
Prevent auto hide labels when redirect panels are rendered.

natocTo:


src/Tracy/assets/Bar/bar.js
```diff
@@ -275,8 +275,9 @@
 	};
 
 	Bar.prototype.autoHideLabels = function() {
-		var labels = this.elem.querySelectorAll('.tracy-label');
-		for (var i = labels.length - 1; i >= 0 && this.elem.clientHeight >= 40; i--) {
+		var labels = this.elem.querySelectorAll('.tracy-label');¨
```
JanTvrdik: typos at the end of line

JanTvrdik:
Unfortunately the problem is a bit more complicated. Each `<ul>` needs to be handled independently.

JanTvrdik:
This may work

~~~css
#tracy-debug-bar ul {
    overflow: hidden;
}
~~~

~~~js
forEach(this.elem.children, function (ul) {
    var labels = ul.querySelectorAll('.tracy-label');
    for (var i = labels.length - 1; i >= 0 && ul.clientHeight >= 40; i--) {
        labels.item(i).hidden = true;
    }
});
~~~

natocTo:
Ok, I am going to change it.

src/Tracy/assets/Bar/bar.js
```diff
@@ -276,9 +276,13 @@
 
 	Bar.prototype.autoHideLabels = function() {
 		var labels = this.elem.querySelectorAll('.tracy-label');
-		for (var i = labels.length - 1; i >= 0 && this.elem.clientHeight >= 40; i--) {
-			labels.item(i).hidden = true;
-		}
+
```
JanTvrdik: this above line is pointless

dg:
Why overflow: hidden?

JanTvrdik:
Because otherwise `ul.clientHeight === 0`. It's based on http://stackoverflow.com/questions/8721001/putting-a-border-around-floating-elements / http://www.quirksmode.org/css/clearing.html

JanTvrdik:
Note that this fixed redirects but not ajax. I have the patch for AJAX I'm just not sure whether we want to include it here.

dg:
Yep, because LI has float:left… It's time to change wordaround to solid solution - flex.

dg:
Now it removes all labels in first row and some labels in second row

dg:
Closed by 46134d3c1af9bf0c57c673c897fb4de418455a3c

PR #194, opened 2016-06-11
Fix for a hidden panel content in a new window.

Korel:
Since fe7df9ab3361329225686108339838b4304d50f6 the content of a Tracy panel has been hidden when you put it in a new window.

This css change returns the origin behaviour.


dg:
Thanks!

PR #198, opened 2016-06-20
ISession added

hranicka:
I run a CLI application in a debug mode (development, CI).

But Tracy v2.4 is starting a session automatically and itselfs: [[1]](https://github.com/nette/tracy/blame/master/src/Bridges/Nette/TracyExtension.php#L101), [[2]](https://github.com/nette/tracy/blame/master/src/Tracy/Debugger.php#L208).

I can't solve this by fake session object. When DI Container is initialized, `session_start()` is called and I do nothing with it.

So I've extracted `session_*` code into `ISession` interface which should allow set custom ISession implementation. It is useful in a CLI when debug mode is enabled.

---

https://forum.nette.org/cs/26250-pojdte-otestovat-nette-2-4-rc?p=3#p175951
http://zlml.cz/fix-compatibility-with-nette-2-4#comment-2740141135

### DIFF
--- a/src/Tracy/Bar.php
+++ b/src/Tracy/Bar.php
@@ -56,7 +56,7 @@ class Bar
 	 */
 	public function render()
 	{
-		$useSession = $this->dispatched && session_status() === PHP_SESSION_ACTIVE;
+		$useSession = $this->dispatched && Debugger::getSession()->isActive();
 		$redirectQueue = & $_SESSION['_tracy']['redirect'];
 
 		if (!Helpers::isHtmlMode() && !Helpers::isAjax()) {
--- a/src/Tracy/BlueScreen.php
+++ b/src/Tracy/BlueScreen.php
@@ -58,7 +58,7 @@ class BlueScreen
 	 */
 	public function render($exception)
 	{
-		if (Helpers::isAjax() && session_status() === PHP_SESSION_ACTIVE) {
+		if (Helpers::isAjax() && Debugger::getSession()->isActive()) {
 			ob_start(function () {});
 			$this->renderTemplate($exception, __DIR__ . '/assets/BlueScreen/content.phtml');
 			$contentId = $_SERVER['HTTP_X_TRACY_AJAX'];
--- a/src/Tracy/Debugger.php
+++ b/src/Tracy/Debugger.php
@@ -120,6 +120,9 @@ class Debugger
 	/** @var ILogger */
 	private static $fireLogger;
 
+	/** @var ISession */
+	private static $session;
+
 
 	/**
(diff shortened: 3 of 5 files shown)


src/Tracy/Bar.php
```diff
@@ -56,7 +56,7 @@ public function getPanel($id)
 	 */
 	public function render()
 	{
-		$useSession = $this->dispatched && session_status() === PHP_SESSION_ACTIVE;
+		$useSession = $this->dispatched && Debugger::getSession()->isActive();
```
hranicka: Maybe this file should be untouched? (`session_status() === PHP_SESSION_ACTIVE`)

src/Tracy/BlueScreen.php
```diff
@@ -58,7 +58,7 @@ public function addPanel($panel)
 	 */
 	public function render($exception)
 	{
-		if (Helpers::isAjax() && session_status() === PHP_SESSION_ACTIVE) {
+		if (Helpers::isAjax() && Debugger::getSession()->isActive()) {
```
hranicka: Same question as above (keep `session_status() === PHP_SESSION_ACTIVE`)?

dg:
Simply don't call `Tracy\Debugger::dispatch` in `TracyExtension` in CLI, this is unnecessary complicated solution.

hranicka:
@dg But `TracyExtension` adds `Tracy\Debugger::dispatch()` call automatically in `Container::initialize`, see https://github.com/nette/tracy/blame/master/src/Bridges/Nette/TracyExtension.php#L101

This is my problem. I have fake session, but with v2.4 it doesn't work correctly because Tracy checks for `session_status` and calls `session_starts` internally.

hranicka:
@dg Or did you meant throw away these changes (ISession inside Debugger) and rather edit `TracyExtension` and allow/disallow [this line](https://github.com/nette/tracy/blame/master/src/Bridges/Nette/TracyExtension.php#L101) call?

dg:
Yes :-)

hranicka:
Hmm, and is explicit call of `Tracy\Debugger::dispatch();` really needed in Container::initialize?

What about if I replace

```
$initialize->addBody('if ($tmp = $this->getByType("Nette\Http\Session", FALSE)) { $tmp->start(); Tracy\Debugger::dispatch(); };');
```

with


```
$initialize->addBody('if ($tmp = $this->getByType("Nette\Http\Session", FALSE)) { $tmp->start(); };');
```

?

dg:
It is needed for AJAX requests and redirects in bar.

hranicka:
@dg Thank you for your time and explanation. I've made another PR, see #200 .

Closing this.

PR #200, opened 2016-06-20
TracyExtension: Don't dispatch Debugger in a CLI

hranicka:
See #198

I've disabled Session/Debugger dispatching in for CLI SAPI.

It's needed [only for Ajax](https://github.com/nette/tracy/pull/198#issuecomment-227082488).

### DIFF
--- a/src/Bridges/Nette/TracyExtension.php
+++ b/src/Bridges/Nette/TracyExtension.php
@@ -35,10 +35,14 @@ class TracyExtension extends Nette\DI\CompilerExtension
 	/** @var bool */
 	private $debugMode;
 
+	/** @var bool */
+	private $cliMode;
 
-	public function __construct($debugMode = FALSE)
+
+	public function __construct($debugMode = FALSE, $cliMode = FALSE)
 	{
 		$this->debugMode = $debugMode;
+		$this->cliMode = $cliMode;
 	}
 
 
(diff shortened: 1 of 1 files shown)


dg:
Thanks!

hranicka:
This is the simplest possible solution, thanks @dg for your mentoring.

JanTvrdik:
PHP_SAPI is not part of DI container cache key

hranicka:
@JanTvrdik I know, but this is same as https://github.com/nette/http/blob/master/src/Bridges/HttpDI/HttpExtension.php#L57 or not?

dg:
It is "coincidentally" part of cache key, the question is whether make it official part of key, or modify implementation or leave it as coincidentally part of key.

hranicka:
There is also a parameter `consoleMode` [here](https://github.com/nette/bootstrap/blob/master/src/Bootstrap/Configurator.php#L145).

Does it have any impact on it?

I can change this something like:

```php

$initialize->addBody('if (PHP_SAPI !== 'cli' && $tmp = $this->getByType("Nette\Http\Session", FALSE)) { $tmp->start(); Tracy\Debugger::dispatch(); };');

```

JanTvrdik:
@dg I think that it should be possible to generate DI container for application from CLI.

dg:
It would actually be parameter for TracyExtension, like $debugMode .

dg:
@hranicka can you change it this way? https://github.com/nette/http/commit/27c3de5248358b494bf9af557d800e7d3e1d53e9#diff-e18d2abacdd85c4cc65b1a8d611ca31f

hranicka:
Ok, I edit it.

So Tracy after this change won't be compatible with the current Bootstrap.

I've tried use config parameters with `%consoleMode%` but I must add more code because of [this magic](https://github.com/nette/tracy/blob/master/src/Bridges/Nette/TracyExtension.php#L76-L84). 

Constructor parameters seems better.

---

And what about Nette\Http packages? There is also PHP_SAPI constant.

dg:
:+1:

hranicka:
I can prepare you PR into nette/bootstrap for your changes in nette/http and tracy. Give me a few minutes.

dg:
I have it finished.

hranicka:
Ok, thanks :)

PR #207, opened 2016-07-14
All H1 tags should make panel draggable

kukulich:


kukulich:
Why not? We have one panel with two sections, both with heading.

PR #210, opened 2016-07-25
WIP: Notifiers for Logger, separate MailNotifier class

finwe:
This is an initial commit for this long forsaken RFC https://forum.nette.org/en/24668-notifiers-for-tracy-debugger

Needed to be (determined how to be) solved:

[ ] - Deprecation messages for mail related public Logger properties
[ ] - BC Break - $mailer is now moved to MailNotifier
[ ] - Dependency of MailNotifier on Logger
[ ] - Tests

Thanks for comments and suggestions

### DIFF
--- a/src/Tracy/INotifier.php
+++ b/src/Tracy/INotifier.php
@@ -1,0 +1,19 @@
+<?php
+
+/**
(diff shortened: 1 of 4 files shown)


dg:
Related to #166

PR #213, opened 2016-07-28
Fixed assets path if session path is different

MartkCz:
Fixed: https://forum.nette.org/cs/26772-problem-s-cookie-path-podle-url


dg:
Thanks

PR #215, opened 2016-08-05
Fix implode args order

JanTvrdik:
https://github.com/nette/tracy/blob/master/src/Tracy/Logger.php#L94 (works only because of BC)


JanTvrdik:
Converted to PR.

Issue #216, opened 2016-08-09
Debug bar breaks Xdebug session when redirect occurs


richard-ejem:
Scenario:
* use Tracy\Debugger::enable() and place it after a possible redirect (i.e. to a language-specific url, like http://example.com/ -> http://example.com/en; should not happen in Nette application with standard bootstrap, happened to me in Drupal project using Tracy)
* the redirect occurs when loading Tracy assets (like http://example.com/?_tracy_bar=js&v=2.4.2&XDEBUG_SESSION_STOP=1&XDEBUG_PROFILE=0&XDEBUG_TRACE=0 is redirected to http://example.com/en?_tracy_bar=js&v=2.4.2&XDEBUG_SESSION_STOP=1...
* problem: header_remove('Set-Cookie') is not called from Bar.php when the redirect occurs and the cookie XDEBUG_SESSION=deleted gets into browser

I've proposed a simple solution printing at least console warning to confused developers like me, who would be investigating for hours why XDEBUG stops after first request, will pull request it in a second :)

richard-ejem:
Thanks to @JanTvrdik for commenting original pull request, I just found even simpler way to detect and fixed the pull request with a single simple commit :)

dg:
What about this 6a24bb67ce63a072c81a906b13014131e6458fd2?

richard-ejem:
@dg unfortunatelly not helping, tested on my project :( I will provide some minimal case to reproduce the problem soon

dg:
It closes XDEBUG session, but it should close it only once.

richard-ejem:
Take a look at this MWE: https://github.com/richard-ejem/tracytest
try debugging it in browser, debugger stops working after removing the XDEBUG_SESSION_START argument from URL.

dg:
I see, dispatchContent() is not loaded…

richard-ejem:
ok so maybe please re-open the issue?
My suggested fix was just to notify users about the problem, as any redirect before Debugger::enable() when XDEBUG_SESSION_STOP is passed will cause this problem.
Just checked sources of Xdebug, there seems to be no way to stop Xdebug for a single request without destroying the cookie.
Other solution would be to create pull request for Xdebug to add something like XDEBUG_SKIP=1 which would disable Xdebug just for the single request, but it will take long to be merged and downloaded by most users.

richard-ejem:
But there is still one edge case where it can cause problems, see current https://github.com/richard-ejem/tracytest .
In short:

[...]

PR #217, opened 2016-08-09
warn user when redirect occurs while loading Tracy assets

richard-ejem:
- feature (bug-preventing one :) )
- issues - #216
- documentation - not needed
- BC break - no

### DIFF
--- a/src/Tracy/Bar.php
+++ b/src/Tracy/Bar.php
@@ -184,6 +184,7 @@ class Bar
 			header('Cache-Control: max-age=864000');
 			header_remove('Pragma');
 			header_remove('Set-Cookie');
+			require __DIR__ . '/assets/Bar/bar.check-redirect.phpjs';
 			readfile(__DIR__ . '/assets/Bar/bar.js');
 			readfile(__DIR__ . '/assets/Toggle/toggle.js');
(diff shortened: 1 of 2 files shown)


src/Tracy/Bar.php
```diff
@@ -222,4 +214,34 @@ public function dispatchContent()
...
+	{
+		return extension_loaded('xdebug')
+			? ['XDEBUG_SESSION_STOP' => 1, 'XDEBUG_PROFILE' => 0, 'XDEBUG_TRACE' => 0] : [];
+	}
+
+	private function getTracyDebugScriptPath($path)
+	{
+		return htmlspecialchars("$path?")
+		. http_build_query(['_tracy_bar' => 'js', 'v' => Debugger::VERSION] + $this->getStopXdebug(), NULL, '&amp;');
+	}
```
JanTvrdik: I have no idea what have you changed because you combined refactoring and maybe some hidden change into single commit.

f3l1x:
@dg @JanTvrdik What do you think?

dg:
@f3l1x fixed by b0845f5a9ef014fdbdd6bcc912a295bdf7fa87e1

PR #218, opened 2016-08-25
Dumper: default options in static vars

pkristian:
When some SQL fail, i needed to show whole query, so: need to configure truncate length in `\Tracy\Dumper::toHtml()` , so i did everything configurable

### DIFF
--- a/src/Tracy/Dumper.php
+++ b/src/Tracy/Dumper.php
@@ -29,6 +29,21 @@ class Dumper
 		LOCATION_LINK = 0b0010, // appends clickable anchor
 		LOCATION_CLASS = 0b0100; // shows where class is defined
 
+	/** @var int  how many nested levels of array/object properties display {@link Dumper::toHtml()} */
+	public static $depth = 4;
+
+	/** @var int  how truncate long strings? {@link Dumper::toHtml()} */
+	public static $truncate = 150;
+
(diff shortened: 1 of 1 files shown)


dg:
Static vars are antipattern…

janedbal:
I would love to have configurable these options.

pkristian:
Similar variables are in `\Tracy\Debugger`, in fact i copied and altered it

dg:
They are configurable, all methods accepts options.

pkristian:
If there is another , more pleasing way, how to configure configure truncate- mainly for bluesreen purposes, i will be happy, if you will show me

dg:
The way is to add $dumperOptions to BlueScreen.

pkristian:
Doesnt sound bad, but wouldn't be nice to have this more global?

in BlueScreen, `public static $dumperOptions = [];` is fine, or how will you recommand it?

dg:
Without static, you can access it via `Debugger::getBlueScreen()->dumperOptions`

pkristian:
Sounds good, going on it

pkristian:
oh, i just found in v 2.4.2 is nice thing ´\Tracy\Debugger::getBlueScreen()->maxLength´, just what i needed... so thx, David! :)

PR #223, opened 2016-09-18
optimize Dumper::encodeString for long strings

schlndh:
- feature
- issues - none
- documentation - not needed
- BC break - yes

The purpose of this PR is to speedup dumping of large strings for reasonable maxLength. The performance problem is in `preg_match('#[^\x09\x0A\x0D\x20-\x7E\xA0-\x{10FFFF}]#u', $s)` which scans the whole string. By preferring `mb_substr` it is possible to truncate the string before running it through the regex and thus potentially saving a lot of time for large strings. However this is a slight BC break, because now it won't recognize some binary strings it did recognize before, but since the binary part will be outside the truncated string I think it is worth it.

As you can see from [the benchmark I made](https://gist.github.com/schlndh/e7f3e9c7f98ce8131ef3452b8bb46fb5) the performance impact for short strings or large strings with unlimited maxLength is minimal or none, however the performance gain for long strings with small maxLength is massive.

The benchmark compares current `encodeString` implementation (encodeStringOrig) with implementation from this PR (encodeStringChanged) and just running the `preg_match` on the whole string. The first column (N) is the number of characters and the values of other columns are average times for one call on given string in nanoseconds.

Benchmarking was done on Linux with PHP 7.0.10

### DIFF
--- a/src/Tracy/Dumper.php
+++ b/src/Tracy/Dumper.php
@@ -428,7 +428,7 @@ class Dumper
 			}
 			$s = strtr($s, $table);
 
-		} elseif ($maxLength && $s !== '' && !function_exists('mb_substr')) {
+		} elseif ($maxLength && strlen($s) > $maxLength && !function_exists('mb_substr')) {
 			$i = $len = 0;
 			do {
 				if (($s[$i] < "\x80" || $s[$i] >= "\xC0") && (++$len > $maxLength)) {
--- a/src/Tracy/Dumper.php
+++ b/src/Tracy/Dumper.php
@@ -415,26 +415,31 @@ class Dumper
 			$table["\t"] = '\t';
 		}
 
-		if (preg_match('#[^\x09\x0A\x0D\x20-\x7E\xA0-\x{10FFFF}]#u', $s) || preg_last_error()) {
+		if (function_exists('mb_substr')) {
+			if ($maxLength && strlen($s) > $maxLength) {
+				$s = mb_substr($tmp = $s, 0, $maxLength, 'UTF-8');
+				$shortened = $s !== $tmp;
+			} else {
+				$shortened = false;
+			}
+			if (preg_match('#[^\x09\x0A\x0D\x20-\x7E\xA0-\x{10FFFF}]#u', $s) || preg_last_error()) {
+				$s = strtr($s, $table);
+			}
+		} elseif (preg_match('#[^\x09\x0A\x0D\x20-\x7E\xA0-\x{10FFFF}]#u', $s) || preg_last_error()) {
 			if ($shortened = ($maxLength && strlen($s) > $maxLength)) {
 				$s = substr($s, 0, $maxLength);
 			}
 			$s = strtr($s, $table);
 
 		} elseif ($maxLength && $s !== '') {
-			if (function_exists('mb_substr')) {
-				$s = mb_substr($tmp = $s, 0, $maxLength, 'UTF-8');
-				$shortened = $s !== $tmp;
-			} else {
-				$i = $len = 0;
-				do {
-					if (($s[$i] < "\x80" || $s[$i] >= "\xC0") && (++$len > $maxLength)) {
-						$s = substr($s, 0, $i);
-						$shortened = TRUE;
-						break;
-					}
-				} while (isset($s[++$i]));
-			}
+			$i = $len = 0;
+			do {
+				if (($s[$i] < "\x80" || $s[$i] >= "\xC0") && (++$len > $maxLength)) {
+					$s = substr($s, 0, $i);
+					$shortened = TRUE;
+					break;
+				}
+			} while (isset($s[++$i]));
 		}
 
 		return $s . (empty($shortened) ? '' : ' ... ');
--- a/src/Tracy/Dumper.php
+++ b/src/Tracy/Dumper.php
@@ -415,23 +415,20 @@ class Dumper
 			$table["\t"] = '\t';
 		}
 
-		if (function_exists('mb_substr')) {
+		$shortened = FALSE;
+		if ($maxLength && function_exists('mb_substr') && strlen($s) > $maxLength) {
+			$s = mb_substr($tmp = $s, 0, $maxLength, 'UTF-8');
+			$shortened = $s !== $tmp;
+		}
+
+		if (preg_match('#[^\x09\x0A\x0D\x20-\x7E\xA0-\x{10FFFF}]#u', $s) || preg_last_error()) {
 			if ($maxLength && strlen($s) > $maxLength) {
-				$s = mb_substr($tmp = $s, 0, $maxLength, 'UTF-8');
-				$shortened = $s !== $tmp;
-			} else {
-				$shortened = false;
-			}
-			if (preg_match('#[^\x09\x0A\x0D\x20-\x7E\xA0-\x{10FFFF}]#u', $s) || preg_last_error()) {
-				$s = strtr($s, $table);
-			}
-		} elseif (preg_match('#[^\x09\x0A\x0D\x20-\x7E\xA0-\x{10FFFF}]#u', $s) || preg_last_error()) {
-			if ($shortened = ($maxLength && strlen($s) > $maxLength)) {
 				$s = substr($s, 0, $maxLength);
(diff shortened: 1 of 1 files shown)


JanTvrdik:
How will `mb_substr` handle binary (not UTF-8) strings?

src/Tracy/Dumper.php
```diff
@@ -415,26 +415,31 @@ public static function encodeString($s, $maxLength = NULL)
...
-				} while (isset($s[++$i]));
-			}
+			$i = $len = 0;
+			do {
+				if (($s[$i] < "\x80" || $s[$i] >= "\xC0") && (++$len > $maxLength)) {
+					$s = substr($s, 0, $i);
+					$shortened = TRUE;
+					break;
+				}
+			} while (isset($s[++$i]));
```
JanTvrdik: This code block should be only done when the string was not already shortened by mb_substr, i.e. 

~~~php
} elseif ($maxLength && $s !== '') {
~~~

should be changed to 

~~~php
} elseif ($maxLength && $s !== '' && !function_exists('mb_substr')) {
~~~

src/Tracy/Dumper.php
```diff
@@ -415,26 +415,31 @@ public static function encodeString($s, $maxLength = NULL)
...
+			if ($maxLength && strlen($s) > $maxLength) {
+				$s = mb_substr($tmp = $s, 0, $maxLength, 'UTF-8');
+				$shortened = $s !== $tmp;
+			} else {
+				$shortened = false;
+			}
+			if (preg_match('#[^\x09\x0A\x0D\x20-\x7E\xA0-\x{10FFFF}]#u', $s) || preg_last_error()) {
+				$s = strtr($s, $table);
+			}
+		} elseif (preg_match('#[^\x09\x0A\x0D\x20-\x7E\xA0-\x{10FFFF}]#u', $s) || preg_last_error()) {
```
JanTvrdik: The speed of the `preg_match` call be made faster by inverting the regexp, i.e. changing

~~~php
preg_match('#[^\x09\x0A\x0D\x20-\x7E\xA0-\x{10FFFF}]#u', $s) || preg_last_error()
~~~

to

~~~php
!preg_match('#^[\x09\x0A\x0D\x20-\x7E\xA0-\x{10FFFF}]*+\z#u', $s) || preg_last_error()
~~~

src/Tracy/Dumper.php
```diff
@@ -415,26 +415,31 @@ public static function encodeString($s, $maxLength = NULL)
...
+		if (function_exists('mb_substr')) {
+			if ($maxLength && strlen($s) > $maxLength) {
+				$s = mb_substr($tmp = $s, 0, $maxLength, 'UTF-8');
+				$shortened = $s !== $tmp;
+			} else {
+				$shortened = false;
+			}
+			if (preg_match('#[^\x09\x0A\x0D\x20-\x7E\xA0-\x{10FFFF}]#u', $s) || preg_last_error()) {
+				$s = strtr($s, $table);
+			}
```
JanTvrdik: This code duplication can be avoided if you change the following `elseif ` back to `if`

src/Tracy/Dumper.php
```diff
@@ -415,26 +415,31 @@ public static function encodeString($s, $maxLength = NULL)
...
+		} elseif (preg_match('#[^\x09\x0A\x0D\x20-\x7E\xA0-\x{10FFFF}]#u', $s) || preg_last_error()) {
 			if ($shortened = ($maxLength && strlen($s) > $maxLength)) {
 				$s = substr($s, 0, $maxLength);
 			}
 			$s = strtr($s, $table);
```
JanTvrdik: After changing the elseif statement back to if, you wil need to change the `$shortened` variable only the string is futher, i.e.

~~~php
if ($maxLength && strlen($s) > $maxLength) {
	$s = substr($s, 0, $maxLength);
	$shortened = TRUE;
}
$s = strtr($s, $table);
~~~

src/Tracy/Dumper.php
```diff
@@ -415,26 +415,31 @@ public static function encodeString($s, $maxLength = NULL)
...
 		}
 
-		if (preg_match('#[^\x09\x0A\x0D\x20-\x7E\xA0-\x{10FFFF}]#u', $s) || preg_last_error()) {
+		if (function_exists('mb_substr')) {
+			if ($maxLength && strlen($s) > $maxLength) {
+				$s = mb_substr($tmp = $s, 0, $maxLength, 'UTF-8');
+				$shortened = $s !== $tmp;
+			} else {
+				$shortened = false;
+			}
```
JanTvrdik: This can be simplified by initializing `$shortend` variable first to `FALSE` (note the uppercase letters)

~~~php
$shortened = FALSE;
if ($maxLength && strlen($s) > $maxLength && function_exists('mb_substr')) {
	$s = mb_substr($tmp = $s, 0, $maxLength, 'UTF-8');
	$shortened = $s !== $tmp;
}
~~~

schlndh:
@JanTvrdik Thanks for your comments. It didn't think of checking `mb_substr` with binary strings before, but it sort of works:
```php
$lengths = [];
for ($i < 0; $i < 1000; ++$i) {
	$s = openssl_random_pseudo_bytes(1000);
	@$lengths[strlen(mb_substr($s, 0, 150, 'UTF-8'))]++;
}
ksort($lengths);
foreach ($lengths as $k => $val) {
	echo $k . " " . str_repeat('|', $val) . "\n";
}
```
It makes the substring mostly around 200 bytes long, which is fine I guess.

src/Tracy/Dumper.php
schlndh: It is virtually the same for all test cases in my benchmark except for the test case where the whole string is 1 multibyte character repeated N times which is measurably faster with your regexp.  I don't really understand why that is, but thanks.

schlndh:
I updated [the benchmark](https://gist.github.com/schlndh/e7f3e9c7f98ce8131ef3452b8bb46fb5) with latest changes to this PR.

src/Tracy/Dumper.php
```diff
@@ -415,26 +415,28 @@ public static function encodeString($s, $maxLength = NULL)
...
-				$i = $len = 0;
-				do {
-					if (($s[$i] < "\x80" || $s[$i] >= "\xC0") && (++$len > $maxLength)) {
-						$s = substr($s, 0, $i);
-						$shortened = TRUE;
-						break;
-					}
-				} while (isset($s[++$i]));
-			}
+		} elseif ($maxLength && $s !== '' && !function_exists('mb_substr')) {
```
JanTvrdik: `$s !== ''` part of the condition can be changed to `strlen($s) > $maxLength`

JanTvrdik:
@dg Your implementation is a lot slower for long UTF-8 strings.

dg:
As I understand, *„the performance problem is in `preg_match('#[^\x09\x0A\x0D\x20-\x7E\xA0-\x{10FFFF}]#u', $s)` which scans the whole string“*, my implementation scans only shortened strings. So it should be faster, or not?

JanTvrdik:
See https://gist.github.com/JanTvrdik/5b8f1e22f71bac00301e74a6ce46d387

schlndh:
@dg I don't have time to test it right now, maybe in the evening, but I think the problem is that you still call `preg_match('##u', $s)` on the whole string which can be very long, that's why I truncated it first so that I can be sure that it is reasonably short before calling `preg_match` on it.

dg:
I see, your intention is to remove utf-8 checking at all before shortening.

schlndh:
@dg Yes.

dg:
Repushed, now it should be faster when even mb_substr is disabled.

JanTvrdik:
> Repushed, now it should be faster when even mb_substr is disabled.

Usually, but if you try string `str_repeat("\x80", 1e5)` with `mb_substr` disabled it is still slow. Simple fix is not allow may be to not allow `$i > $maxLength * 4`

dg:
I know, it is acceptable.

JanTvrdik:
Accepting issue which can be fixed with single line (e.g. `$s = substr($s, 0, $maxLength << 2);`) of code is unnecessary

Issue #136, opened 2016-02-15
Contenct-Security-Policy


enumag:
I'm currently experimenting with [Content-Security-Policy](https://en.wikipedia.org/wiki/Content_Security_Policy) and sure enough Tracy doesn't work unless both `'unsafe-inline'` and `'unsafe-eval'` are allowed. I can enable them for development only of course but I'd like to avoid it. Enabling It could easily lead me to a situation where I miss some real CSP violation because of it and push it to production.

dg:
For checking whether you miss some real CSP violation you can use header `Content-Security-Policy-Report-Only`.

enumag:
@dg I'm aware except it will report Tracy violations as well and I didn't find any way to filter them out to keep only the real ones.

dg:
Tracy bar would sometime in the future work without inline and eval #81, while Bluescreen will require unsafe-inline. This can be solved only using CSP `nonce` (in the meantime it may be used by Bar too). But IMHO nonce is not supported by any browser.

enumag:
According to Can I use it [should](http://caniuse.com/#feat=contentsecuritypolicy2) work in Firefox and Chrome but I didn't test it yet.

dg:
So feel free to send PR :-)

spaze:
@dg Thanks for the fix, it's quite clever to fetch the nonce from the response headers!

PR #227, opened 2016-10-01
Tracy requires 'self' in CSP since 2.4

spaze:
Because it sends a separate HTTP request now (see 13f854d4c11221add4697121e183a147732bb13b) it needs to load  JS from the domain itself, too.

Thanks!


spaze:
Also see #136, `nonce` is the way to go, but this doc update should do in the short term.

spaze:
Thanks @dg!

PR #234, opened 2016-11-23
Bar in session: remove undispatched bars after 60 seconds

matej21:


milo:
Not sure it is correct solution, but it probably solve my issue. I can find 2MB - 90MB session files with `_tracy|a:3:{s:8:"redirect";N;s:3:"bar"; .....` junk.

dg:
Queue is limited that there could be a maximum of 5 items, so it shouldn't accumulate data… https://github.com/nette/tracy/blob/cf59d9cceb3a9c1ed9253b864dd2498043bbaae5/src/Tracy/Bar.php#L98
@milo can you try to explore what is wrong?

JanTvrdik:
Is it possible that those 5 items could possible never expire?

milo:
@dg Thanks to pointing this code. It takes some sense to me now, even I cannot reproduce it.

My guess is, I made some huge dump (~20MB) by mistake and I stopped loading in a browser, so item wasn't picked up. Than again and again wasn't picked up. Than I fixed this dump, but data remains in the session, because ordinary traffic is one tracy bar, or two on redirection.

PR #235, opened 2016-11-26
Added Linux version of 'open-in-editor' tool

vlastavesely:


dg:
what about move it to subdirectories like /windows, /linux etc?

vlastavesely:
yes, I had similar idea (renaming files like"install.linux.sh"), but subdirs seems to be better choice.

dg:
Thanks!

PR #239, opened 2017-01-08
Avoid multiple init for bar links

rolandtoth:
- bug fix? yes
- new feature? no
- BC break? no

When the AJAX bar appears, events are re-added for bar links. For example, when clicking on a bar link the corresponding panel appears and disappears because there are two click events on it. This fix adds a data attribute and events are added only if this attribute is not present on the link.


dg:
Thank you, I fixed it a bit differently.

rolandtoth:
Working fine, thanks!

Issue #206, opened 2016-06-27
CSS styles rewrite by website


JanGalek:
Tracy debug bar content is rewrited by styles css of website, for example I have h background orange, and System info h have too orange, it looks bad, here is preview: 
![tracy_css](https://cloud.githubusercontent.com/assets/13171542/16374986/5a4f90b6-3c59-11e6-8870-2cc509cbe647.jpg)

When I will have some time, I can fix it if you want.

dg:
FIx is welcomed.

JanGalek:
Ok, fast test for this is add to style this:

`h1 {
    background: orange !important;
}`

fix will be soon ;)

PR #240, opened 2017-01-11
bar.css, bluescreen.css: fixed leaks of page's default CSS into Tracy

vlastavesely:
- bug fix? yes

I just made Tracy more sexy :) There is [issue opened](https://github.com/nette/tracy/issues/206) that shows headings being affected by CSS property with `!important` set on. I have fixed this particular case (for headings and paragraphs).

Another problem occured on page with styled pseudoelements `:before` or `:after`. On my page, there are styled headings that are decorated by these pseudoelements. Sadly, Tracy rendered these as well and so its layout went broken. Now, after my fixes, any pseudoelements inside Tracy will be hidden by default except cases where it is need for it.

### DIFF
--- a/src/Tracy/assets/Bar/bar.css
+++ b/src/Tracy/assets/Bar/bar.css
@@ -34,6 +34,13 @@ body#tracy-debug { /* for popup window */
 	text-shadow: none;
 }
 
+#tracy-debug:before,
+#tracy-debug:after,
+#tracy-debug *:before,
+#tracy-debug *:after {
+	display: none;
+}
+
 #tracy-debug b,
 #tracy-debug strong {
 	font-weight: bold;
@@ -64,6 +71,8 @@ body#tracy-debug { /* for popup window */
 #tracy-debug .tracy-panel h2,
 #tracy-debug .tracy-panel h3,
 #tracy-debug .tracy-panel p {
+	color: #333 !important;
+	background: transparent !important;
 	margin: .4em 0;
 }
 
@@ -195,7 +204,8 @@ body#tracy-debug { /* for popup window */
 
 #tracy-debug h1 {
 	font: normal normal 23px/1.4 Tahoma, sans-serif;
-	color: #575753;
+	color: #575753 !important;
+	background: transparent !important;
 	margin: -5px -5px 5px;
 	padding: 0 25px 0 5px;
 	max-width: 700px;
--- a/src/Tracy/assets/BlueScreen/bluescreen.css
+++ b/src/Tracy/assets/BlueScreen/bluescreen.css
@@ -25,6 +25,13 @@
 	text-indent: 0;
 }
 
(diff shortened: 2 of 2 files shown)


adrianbj:
On this note, I would really love if bar.css 's 
```
#tracy-debug * {
```
could be changed to:
```
#tracy-debug-bar * {
```

I make this change every time there is an update so I can include ACE editor in a custom panel. I have mentioned this before, but was hoping that if some css changes are about to be made that this could be reconsidered.

Thanks and sorry for hijacking!

vlastavesely:
Since `#tracy-debug-bar` is inside `#tracy-debug` - at least in Tracy that I have installed on this machine (another from that I have sent PR from) - it should be affected also by my change. Am I missing something? What is exactly your problem?

adrianbj:
> Am I missing something? What is exactly your problem?

I just tested with your changes and I still need to make that change of mine which makes things much easier to override. 

My issue is actually kind of the reverse to what you are addressing here - sorry for the confusion - I just wanted to get the CSS discussion going again.

More info here: https://github.com/nette/tracy/pull/151

adrianbj:
Just thought I'd also post the additional CSS rules I am adding (along with that change from `#tracy-debug` to `tracy-debug-bar`

https://github.com/adrianbj/TracyDebugger/blob/865a669a3ddfc0e528413452235e8aefe50e332e/styles.css#L1-L208

The highlighted lines are the ones relevant to all Tracy installs. The rest are only relevant to some of my custom panels. This combination seems to work well, but perhaps your additions here would avoid the need for some of my added rules?

vlastavesely:
So basically, your problem is that Tracy's style is damaging something you've put *inside* the panel? I fixed issues occurring when *outside* page's CSS damages Tracy's design. I believe these are two different things and should be solved separately. Your problem is primarily in the third party library you are using and there it should be fixed. Maybe it can be fixed in Tracy but, as I said, it is different problem.

However, for test purpose, I've just tried to change `#tracy-debug *` to `#tracy-debug-bar *`. There are some differences after I did so. For example, headings `h2` look a little bit different. I do not have time to solve it on my own. If it bothers you, you should prepare your own PR with solution :)

adrianbj:
> So basically, your problem is that Tracy's style is damaging something you've put inside the panel? I fixed issues occurring when outside page's CSS damages Tracy's design. 

Yes, I realize the difference. I was just trying to contribute to improving the CSS conflicts in both directions. Anyway, sorry for hijacking your PR.

dg:
Thanks!

dg:
I merged only :before and :after reset, for the `!important` part I am trying a different approach.

mrtnzlml:
Sending kudos from zlml.cz... :)
screenshot from 2017-01-13 09 35 59
screenshot from 2017-01-13 09 42 12

PR #241, opened 2017-01-13
All css rules are !important [WIP]

dg:
- bug fix? yes   #206
- BC break? maybe…

All Tracy CSS rules are `!important`, so they cannot be rewriten by styles css of website.

### DIFF
--- a/examples/barDump.php
+++ b/examples/barDump.php
@@ -15,6 +15,24 @@ Debugger::enable(Debugger::DETECT, __DIR__ . '/log');
 
 <p>You can dump variables to bar in rightmost bottom egde.</p>
 
+<style>
+h1 { background: orange !important; }
+h1:after {
+content: "Look at this orange box.";
+  background-color: #FFBA10;
+  border-color: black;
+  border-style: dotted;
+}
+
+#tracy-debug h1 {
+	all: initial;
+}
+
+#tracy-debug *:after {
+	all: unset;
+}
+</style>
+
 <?php
 $arr = [10, 20.2, true, null, 'hello', (object) null, []];
 
--- a/src/Tracy/Bar.php
+++ b/src/Tracy/Bar.php
@@ -155,9 +155,12 @@ class Bar
 		foreach ($this->panels as $id => $panel) {
 			$idHtml = preg_replace('#[^a-z0-9]+#i', '-', $id) . $suffix;
 			try {
-				$tab = (string) $panel->getTab();
-				$panelHtml = $tab ? (string) $panel->getPanel() : null;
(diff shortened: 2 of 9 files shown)


JanTvrdik:
I'm afraid that this is going to break a lot of stuff.

dg:
It should not… Anyway it is targeted to the next bigger release.

vlastavesely:
With panels I use, it seems to be OK. But also I think that many panels that use a lot of CSS would have serious problems after this change. For example, for the guy who commented my last PR it could be very problematic. But, as I said in my PR, I think this is not primarily problem of Tracy but the third-party libraries integrated into it.

Please consider to reset `text-transform` also. I it is not uncommon that `H1` are set to `text-transform: uppercase;` and it was first thing that I tried to test with your proposed changes.

dg:
It automatically appends `!important` to every included CSS, so the most of panels should work.

Yes, it completely breaks injection of web styles to Tracy, but it's purpose.

dg:
@vlastavesely done ec3380b7bcdf0c4a0b8c4879c44e8b82db9b20c3

adrianbj:
> For example, for the guy who commented my last PR it could be very problematic.

Being that guy :) I can say that I agree that this is very problematic. 

I have been playing around with this PR and it looks like it makes it impossible to style any custom panels using an external stylesheet. This is something I do with many of my panels. Now the only way to override the !important is to use inline styles also with an important.

I know I keep harping on it, but changing `#tracy-debug * {` to `#tracy-debug-bar * {` makes a huge difference. With the current master of Tracy with this change I can do as much manipulation of panels as I need. While there is still the possibility of styles from the site overriding those in Tracy, I have managed to get around these by re-defining the styles for everything within Tracy (my own stylesheet that I inject) and making them `!important`. It's taken a while to get everything taken care of, but I am pretty happy now that the Tracy debug bar will look the same on all sites.

dg:
The problem is that Tracy (panels and bar) must not accidentally affect by page CSS style.

The only way to do that (until browsers will support "revert":https://developer.mozilla.org/en-US/docs/Web/CSS/revert) it to reset all important CSS properites. For bar, for panels. This PR resets them in very hard way.

On the other side, there is need to create custom styles for custom panels. It is absolutely easy. You just need to add `#tracy-debug` to all selectors. It can be easily done for 3rd party styles with CSS preprocessor, for example with LESS:

```
#tracy-debug {
	@import (less) 'library.css';
}
```

This PR would makes it only slightly harder, it would need in addition to add `!important` for every property. It can be done with regexp `preg_replace('#(?<!!important);#', '!important;', $css)`.

PR #244, opened 2017-01-28
Update TracyExtension.php

tomaskavalek:
Fix missing 'editorMapping' default parameter

- bug fix? yes
- new feature? no
- BC break? no

I add missing default parameter definition, without this definition app returns „Unknown configuration option“.


dg:
Thanks

tomaskavalek:
You're welcome

PR #248, opened 2017-02-19
Helpers: fixed suggestion when calling undefined method of anonymous …

paveljanda:
Tracy\Helpers

- bug fix? yes
- new feature? no
- BC break? no

Fixed suggestion when calling undefined method of anonymous class


paveljanda:
Fixes following error:

```
TypeError: Argument 1 passed to Tracy\Helpers::getSuggestion() must be of the type array, null given
```

dg:
Thanks, I've widened it a bit.

PR #256, opened 2017-05-22
Call custom logger handlers after calling Tracy\Debugger::log()

clovnrian:
- bug fix?/no
- new feature? yes
- BC break? no

I add possibility to add custom handlers which will be called at the end of Tracy\Debugger::log() function. It will allows us to simply add 3rd party loggers e.g. Monolog or other custom applications.


JanTvrdik:
If you need custom log handling, you should just write custom `Tracy\ILogger` implementation. For integration of Tracy & Monolog, see https://github.com/nextras/tracy-monolog-adapter for example.

clovnrian:
Yes it is possibility, but can it be done just in config, without extension? In MonologExtension I see you need set default tracy logger autowiring to false and then, you set the custom logger after compile.

And what if you want preserve default Tracy behavior and add logging to e.g. Monolog?

JanTvrdik:
> And what if you want preserve default Tracy behavior and add logging to e.g. Monolog

You would need to implement some kind of `ChainLogger` which would call multiple `ILogger` implementations.

fprochazka:
I think that if you were to implement the `ChainLogger` for `Tracy`, it would be more helpful than creating new interfaces.
The `ChainLogger` could be the default implementation of the `tracy.logger` service and the extensions could hook themselfs into it (or whatever it would need to do).
Also the configuration could be very simplified.

```yml
services:
    tracy.logger:
        setup:
            - addLogger(@customLogger)
```

or maybe even

```yml
tracy:
    loggers:
        - @customLogger
```

> And what if you want preserve default Tracy behavior and add logging to e.g. Monolog?

[Kdyby/Monolog](https://github.com/Kdyby/Monolog) tries to do that :)

PR #258, opened 2017-05-29
Show date/time error occurred

bumperbox:
- bug fix? no  
- new feature? yes
- BC break? no
- doc PR: not applicable

Often i get emailed screenshots of the 500 error page, if it showed the date/time on the page, it would be much easier to find the relevant logs.

This change does not require any additional tests.


f3l1x:
I like an idea. What do you think @dg? Could we merge it?

dg:
Ok. I'll change it to format ` 22. 6. 2017 11:35`, because `Thu, 22 Jun 2017 11:36:47 +0200` is not understandable outside english speaking countries.

PR #264, opened 2017-10-06
make the error handler context argument optional

xabbuh:
The context argument of an error handler is optional. Furthermore, it will be deprecated in PHP 7.2 (see http://www.php.net/manual/en/function.set-error-handler.php).

- bug fix? yes
- new feature? no
- BC break? no
- doc PR: -

### DIFF
--- a/src/Tracy/Debugger.php
+++ b/src/Tracy/Debugger.php
@@ -361,7 +361,7 @@ class Debugger
 	 * @throws ErrorException
 	 * @internal
 	 */
-	public static function errorHandler($severity, $message, $file, $line, $context)
+	public static function errorHandler($severity, $message, $file, $line, $context = array())
 	{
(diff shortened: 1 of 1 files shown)


src/Tracy/Debugger.php
```diff
@@ -361,7 +361,7 @@ public static function exceptionHandler($exception, $exit = true)
 	 * @throws ErrorException
 	 * @internal
 	 */
-	public static function errorHandler($severity, $message, $file, $line, $context)
+	public static function errorHandler($severity, $message, $file, $line, $context = array())
```
dg: Please change `array()` to `[]`

xabbuh: done

dg:
Thanks!

xabbuh:
Thanks for merging!

PR #276, opened 2017-12-22
Allow adding of custom icons to panel header. By being explicit about what the "window" icon does, we can add others without them triggering this action.

adrianbj:


adrianbj:
Hey @dg - is there any reason not to include this? I have been running it for several months now and it works fine and really helps out - it lets other add other icons to the panels - in my case I added "large window" and "small window" icons so it's easy to expand / contract the panel as needed.

dg:
I just overlooked it.

adrianbj:
Thanks very much!

dg:
Thanks for PR :)

PR #277, opened 2018-01-01
Add support for custom css, js, and body strings

adrianbj:
This provides an easy way to inject dynamically generated css or js (or any other content). This complements the recently added $customCssFiles and $customJsFiles options. The problem with those is that if you use PHP to generate JS/CSS based on conditions, there is no easy way to inject it to affect the debug bar itself (not just specific panels). This is especially difficult (impossible?) when there is a fatal error and the bluescreen is showing.

Thanks for considering. Maybe you have another idea for how to implement similar functionality?

### DIFF
--- a/src/Tracy/Bar.php
+++ b/src/Tracy/Bar.php
@@ -244,7 +244,7 @@ class Bar
 			__DIR__ . '/assets/BlueScreen/bluescreen.js',
 		], Debugger::$customJsFiles));
 
-		echo "(function(){var el = document.createElement('div'); el.className='tracy-debug'; el.innerHTML='".preg_replace('#\s+#u', ' ', Debugger::$customHeadStr)."'; document.head.appendChild(el);})();\n";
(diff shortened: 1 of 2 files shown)


adrianbj:
I have reworked this a bit to have dedicated `$customJsStr` and `$customCssStr` along with a `$customBodyStr` option. This combination seems to work for all my needs. Hopefully others will also find this useful.

adrianbj:
I am guessing you didn't overlook this one as well - I know it might seem a little obscure, but this combination of options makes it really nice for modifying the debug bar and having the modifications still work when the bluescreen is showing which isn't really possible otherwise, especially when you need the JS to be built up with PHP variables injected into it. If you don't like this approach, is there something else you'd consider that might work?

dg:
You right, it looks really obscure and like something that is useful only for you use case. And I do not want to put such things in the code because I will not know why they are there.

And at the same time, I do not want to think about it and look for another solution, sorry…

adrianbj:
Fair enough - thanks for getting back to me anyway!

Issue #280, opened 2018-02-19
Simplify logging configuration for Docker


JanTvrdik:
- bug report? no
- feature request? yes

### Description
To make Tracy work well with Docker, it needs to log to stdout or stderr. This is currently possible, but not easy to do. In this issue I'd like to discuss how we can make this common use-case easier.

The logger needs to be configured before or during `Tracy::enable()` call, otherwise errors which occur after `Tracy::enable()` and before `Tracy::setLogger()` will not be logged to stdout/stderr. Logging of error which occur before `Tracy::enable()` is handled by PHP itself (by configuring `error_log` and `log_errors` php.ini options).

Current solution may look like this

~~~php
Tracy\Debugger::setLogger(new class extends Tracy\Logger {
	public function __construct() {
		// intentionally do not call parent constructor,
		// because we don't actually need the parameters
	}


	public function log($value, $priority = self::INFO) {
		@file_put_contents(
			'php://stderr',
			str_replace("\n", " ", $this->formatMessage($value)) . "\n",
			FILE_APPEND | LOCK_EX
		);
	}
});

Tracy\Debugger::enable();
~~~

This is solution is ugly (we don't use most of Tracy\Logger, only the `formatMessage` method) and way to long to write.

### Proposed Solution A

~~~php
// specify path to file instead of log directory
Tracy\Debugger::enable(Tracy\Debugger::DETECT, 'php://stderr');
~~~

### Proposed Solution B

~~~php
// allow Tracy\ILogger instance as alternative to "log directory"
// and define and add Tracy\StreamLogger class
$logger = new Tracy\StreamLogger('php://stderr');
Tracy\Debugger::enable(Tracy\Debugger::DETECT, $logger);
~~~

---

Note that in both cases, generating bluescreens will be disabled. This is not a problem for me, because after DI container is successfully compiled and initialized, a smarter (and more complex) logger will be used instead which handles this. But maybe other people have different requirements and use-cases so that should be probably considered as well.

f3l1x:
[...]

PR #281, opened 2018-02-20
Logger refactoring

JanTvrdik:
- bug fix? no
- new feature? yes
- BC break? no

Implements https://github.com/nette/tracy/issues/280 with lots of hack to maintain compatibility.


JanTvrdik:
Not sure if the hacks are worth it. It may be better to only add StreamLogger and allow logger instance as alternative to log directory but do not modify the Logger at all.

And possible release the refactored logger in the future with BC breaks as part of Tracy 3.

src/Tracy/Logger.php
```diff
@@ -117,87 +132,42 @@ protected function formatLogLine($message, $exceptionFile = null)
...
-			? $this->emailSnooze
-			: @strtotime($this->emailSnooze) - time(); // @ timezone may not be set
-
-		if ($this->email && $this->mailer
-			&& @filemtime($this->directory . '/email-sent') + $snooze < time() // @ file may not exist
-			&& @file_put_contents($this->directory . '/email-sent', 'sent') // @ file may not be writable
-		) {
-			call_user_func($this->mailer, $message, implode(', ', (array) $this->email));
-		}
+		$reflection = new \ReflectionMethod($this->mailLogger, 'sendEmail');
```
f3l1x: Why is reflection needed? If MailLogger is not interface, it might have sendEmail public. WDYT?

f3l1x:
I like this PR, maybe it could be more cleaner (without hacks) for nette 3.0? If we break the API, let's make better one.

PR #282, opened 2018-02-21
Added support for __debugInfo() magic method.

sallyx:
- bug fix? no
- new feature? yes
- BC break? yes

After this change result from \Tracy\Dumper::dump() will reflect the return value from __debugInfo() magic method (if defined).

As this might change output, I consider it as a BC break.

### DIFF
--- a/src/Tracy/Dumper.php
+++ b/src/Tracy/Dumper.php
@@ -7,8 +7,6 @@
 
 namespace Tracy;
 
-use Tracy;
-
 
 /**
  * Dumps a variable.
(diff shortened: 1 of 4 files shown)


JanTvrdik:
I'm strongly against using `__debugInfo()` by default because it hides the actual object state.

JanTvrdik:
Also see https://github.com/nette/tracy/pull/135, you can use

~~~php
Tracy\Dumper::$objectExporters[NULL] = function ($obj) {
    return method_exists($obj, '__debugInfo') ? $obj->__debugInfo() : (array) $obj;
};
~~~

sallyx:
Yes, but that's exactly what is __debugInfo() for :-) 

I missed that PR. Anyway, thanks for the code example ;-)

milo:
@sallyx I trust to my code, so I would accept `__debugInfo()` on my classes. But when I work with 3rd party code, I really want to see the real content of the object and not the trimmed one.

sallyx:
Ok, I could add option [ Dumper::DEBUGINFO => FALSE ] like it was in #135 ...

PR #283, opened 2018-02-22
StreamLogger

JanTvrdik:
- bug fix? no
- new feature? yes
- BC break? no

implements #280

### DIFF
--- a/src/Tracy/Debugger.php
+++ b/src/Tracy/Debugger.php
@@ -138,12 +138,12 @@ class Debugger
 
 	/**
 	 * Enables displaying or logging errors and exceptions.
-	 * @param  mixed   production, development mode, autodetection or IP address(es) whitelist.
-	 * @param  string  error log directory
-	 * @param  string  administrator email; enables email sending in production mode
+	 * @param  mixed           production, development mode, autodetection or IP address(es) whitelist.
+	 * @param  string|ILogger  error log directory or logger instance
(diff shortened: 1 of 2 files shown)


dg:
Isn't better this

```php
Debugger::setLogger(new StreamLogger);
Debugger::enable();
```

than this?

```php
Debugger::enable(Debugger::DETECT, new StreamLogger);
```

JanTvrdik:
The issue is that in Nette you're never actually calling `Debugger::enable()`, because there is `$configurator->enableDebugger()`. So, the actual comparison is

~~~php
Tracy\Debugger::setLogger(new Tracy\StreamLogger('php://stderr'));

$configurator = new Nette\Configurator();
$configurator->enableDebugger();
~~~

vs.

~~~php
$configurator = new Nette\Configurator();
$configurator->enableDebugger(new Tracy\StreamLogger('php://stderr'));
~~~

JanTvrdik:
When I think about it now, the logic could be moved to `Nette\Configurator` – but it would create disparity between `Tracy\Debuger::enable()` and `Nette\Configurator::enableDebugger()` signatures.

dg:
> create disparity between

It does not matter.

dg:
@JanTvrdik Do you want to move it somehow in v2.5? Or deprecate something, so the next version does not have to require a lot of hacks?

PR #284, opened 2018-03-11
Get nonce from both CSP and CSP-Report-Only headers

PavelJurasek:
- bug fix? no
- new feature? yes
- BC break? no
- doc PR: will do

This one is also related to https://github.com/nette/http/pull/135.


dg:
So slow!!! :-)

PR #287, opened 2018-03-30
info.panel.html: fix undefined property in list of composer packages

vlastavesely:
- bug fix? yes
- BC break? no

Field `source` is not mandatory. For example, my local repositories
mirrored from a path do not have the field in `composer.lock`. Refs
to `source` and `dist` are the same on all my dependencies and even
in file `Locker.php` of `composer/composer` package there is the
retrieval of reference ID fall-backed by using `dist` field.

https://github.com/composer/composer/blob/28e41d094e78e0a60dc5123ee572701154008848/src/Composer/Package/Locker.php#L417


dg:
Thanks

Issue #290, opened 2018-04-11
position:fixed causing z-index issues


adrianbj:
- bug report? yes
- feature request? no
- version: 2.5-dev

### Description
Because of:
```
#tracy-debug:not(body) {
	position: fixed;
```
debug bar panels are no longer appearing on top of any element that has a `z-index` set (even if it's very low).

Changing position back to absolute immediately fixes the panels, but of course it has the side-effect that the debug bar is no longer sticky to the bottom of the page when scrolling.

Not yet sure of the optimal solution to fix this at the moment, but it's pretty critical to fix.

PR #291, opened 2018-04-11
Attempted fix for panels not appearing above other elements with defined z-index

adrianbj:
This seems to fix https://github.com/nette/tracy/issues/290 although it isn't well tested and it mostly just undoes some of your recent changes. Also not sure about the calc - 100px but otherwise the horizontal scrollbar appears for the page.


dg:
Thanks!

PR #296, opened 2018-04-30
Add Psalm to Tracy build, fixing issues it found

muglug:
- bug fix? no
- new feature? no
- BC break? no

This adds Psalm's static analysis to tracy builds, and fixes a bunch of docblocks. It also makes some logic in `src/Tracy/Dumper.php` more explicit so that it can be understood by Psalm and other static analysers.

A few issues are ignored (ones I don't think you care about) but this gets the bulk of them.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,6 +2,9 @@ module.exports = {
 	"env": {
 		"browser": true
 	},
+	"parserOptions": {
+		"ecmaVersion": 6
+	},
 	"globals": {
 		"Tracy": true
 	},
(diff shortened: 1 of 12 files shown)


dg:
Thanks @muglug.

I merged the most of PR here https://github.com/nette/tracy/commit/cd1d45eaeaf33f67693aa74aa1d27456b0b95332 and https://github.com/nette/tracy/commit/b21020b68f76721d5dc52d6c7d0ba527830a85a5

I do not want to add Psalm to travis yet because I do not know it and I want to try it first.

dg:
Thanks :)

PR #301, opened 2018-05-16
Fixed undeclared propery error

### LINKED ISSUE #300: Cannot read an undeclared property Tracy\Bridges\Nette\TracyExtension::$builder
I found a bug after recent Tracy update on Nette 2.4.4. I can't even start application

obrazek

bezdutchek:
fixed accessing $this->builder property which doesn`t exist, replaced with use of correct local variable $builder

- bug fix #300
- BC break? yes


dg:
Thanks

PR #302, opened 2018-05-17
Open-editor windows script - fix url match regex

natocTo:
- bug fix: yes
- BC break: no

Search and replace parameter can be empty when user click to path for open the file.


dg:
Thanks

PR #303, opened 2018-05-18
Linux "open-editor" supports "create" and "fix" actions.

smuuf:
- New feature.
- No BC break intended.
- Based on: https://twitter.com/geekovo/status/997538920278773761

Updated Linux "open-editor" bash script to support the new *"create"* and *"fix"* actions that can now be used via the `editor://` protocol.

- Traditional **`open` action**:
  - Should behave the same as always.
- New **`create` action**:
  - Will create file at the location specified in the `file` parameter, **if it does not already exist**. Then opens the file.
  - Any necessary but non-existing directories are created (recursively).
- New **`fix` action**:
  - Performs search (`search` param) and replace (`replace` param) in the file (`file` param) at the specified line (`line` param, if specified). Then opens the file.
  - Fails if the file does not exist.
- The order of GET parameters in the `editor:// ...` URI is now *not* important - **parameters can be set in any order**.
- Added `$editor` command template for Visual Studio Code.
- Exit with message if the `$editor` variable is not specified.

### DIFF
--- a/tools/open-in-editor/linux/open-editor.sh
+++ b/tools/open-in-editor/linux/open-editor.sh
@@ -1,39 +1,101 @@
 #!/bin/bash
+declare -A mapping
 
+#
+# Configure your editor by setting the $editor variable:
+#
+
+# Visual Studio Code
+#editor='code --goto "$FILE":"$LINE"'
 # Emacs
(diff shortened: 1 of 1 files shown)


dg:
Wow, thank you so much!

PR #309, opened 2018-07-09
Helpers: take host name from SERVER_NAME instead of HTTP_HOST

adaamz:
bug fix
BC break? yes

Hi,
it is better to identify server host name by secure directive SERVER_NAME, because HTTP_HOST is not secure and can be changed by user.
https://www.skeletonscribe.net/2013/05/practical-http-host-header-attacks.html


dg:
Why do you think SERVER_NAME can't be spoofed in the same way?

adaamz:
@dg https://www.geeksforgeeks.org/what-is-the-difference-between-http_host-and-server_name-in-php/
SERVER_NAME should be obtained from server configuration (hardcoded)
HTTP_HOST is obtained from HTTP headers (which can anybody change)

dg:
The question should have been different. Why I'd like to see SERVER_NAME instead of spoofed HTTP_HOST in source line? HTTP_HOST better reflects what is in the address bar.

PR #313, opened 2018-08-20
fix path to composer.lock in info.panel.phtml

zdenekgebauer:
fixes warning about incorrect path to composer.lock when folder "vendor" is not placed in root

- bug fix / new feature?   
- BC break? yes/no
- doc PR: nette/docs#???  

Line $composer = @json_decode(file_get_contents($lockFile)); throws warning about incorrect $lockFile when composer.json contains  config "vendor-dir": "some/subdir".
I use workaround with $_SERVER['DOCUMENT_ROOT']. Is there any better solution how to retrieve composer.lock?


dg:
fixed 9be5b49e0991f45c348a8966b69e4e7dd2c27caf

PR #314, opened 2018-08-27
PSR-3 adapter

JanTvrdik:
@dg Would you accept PR with PSR-3 adapter? So it would be possible to use any PSR-3 compatible logger (e.g. Monolog) with Tracy, i.e. the following would be possible

~~~php
$psrLogger = new Monolog\Logger(...);
$tracyLogger = new Tracy\Bridges\Psr\PsrLoggerAdapter($psrLogger);
Tracy\Debugger::setLogger($tracyLogger);
~~~

Or we could even apply the adapter automatically, so the following would be possible:

~~~php
$psrLogger = new Monolog\Logger(...);
Tracy\Debugger::setLogger($psrLogger);
~~~

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,8 @@
 	},
 	"require-dev": {
 		"nette/di": "^2.4 || ^3.0",
-		"nette/tester": "^2.0"
+		"nette/tester": "^2.0",
+		"psr/log": "^1.0"
 	},
 	"suggest": {
 		"https://nette.org/donate": "Please support Tracy via a donation"
--- a/src/Bridges/Psr/PsrToTracyLoggerAdapter.php
+++ b/src/Bridges/Psr/PsrToTracyLoggerAdapter.php
@@ -1,0 +1,62 @@
+<?php
+
+/**
+ * This file is part of the Tracy (https://tracy.nette.org)
+ * Copyright (c) 2004 David Grudl (https://davidgrudl.com)
+ */
+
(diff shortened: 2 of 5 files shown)


JanTvrdik:
The issue has been converted to PR with implementation. The tests [failed on Travis](https://travis-ci.org/nette/tracy/jobs/421207087) with and odd error 

> Return value of Tester\Runner\Job::getHeaders() must be of the type array, null returned

which looks like either Tester (cc @milo) or PHP bug.

---

> Is it somehow related to #281 & #283 ?

Not directly, but indirectly, yes.

* `PsrToTracyLoggerAdapter` makes it easy to use any third party PSR-3 logger with Tracy, therefore reducing the need for having `StreamLogger` in Tracy.
* `TracyToPsrLoggerAdapter` makes it easy to use any `Tracy\ILogger` with any library that can log to PSR-3 logger. This makes splitting `Tracy\Logger` to multiple smaller `Tracy\ILogger` implementations even more valuable as it allows for example using `Tracy\BlueScreenLogger` as part one of Monolog handlers.

JanTvrdik:
I restarted the build on Travis by using close-reopen combo and the test now works.

f3l1x:
Cool. 🤗

dg:
Thanks, merged.

slischka:
When will be this released ?

mabar:
Nette 3.0

PR #322, opened 2018-10-03
bar.js: send with Content-type: application/javascript

jakubboucek:
- Type: bug fix
- BC break: yes, but probably no effective
- No tests or docs affected or applied

Tracy is sending Tracy\Bar components with `Content-Type: text/javascript`.

According to [RFC 4329](https://www.rfc-editor.org/rfc/rfc4329.txt) (chapter 7.1) is `text/javascript` **obsolete** and should be used `application/javascript` instead.

In real case it make light impediment to right detect, and process response. For example [`.htaccess` in nette/web-project](https://github.com/nette/web-project/blob/7132061169a4f1f66c443a58e06ccc26b6964807/www/.htaccess#L26) is set `DEFLATE` filter only for `application/javascript` mime. I mean it's just mistake, not intentionally.

### DIFF
--- a/src/Tracy/Bar.php
+++ b/src/Tracy/Bar.php
@@ -174,7 +174,7 @@ class Bar
 	{
 		$asset = $_GET['_tracy_bar'] ?? null;
 		if ($asset === 'js') {
-			header('Content-Type: text/javascript');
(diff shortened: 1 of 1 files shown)


dg:
Thanks

Issue #279, opened 2018-02-02
Option to disable Tracy shutdownHandler


adrianbj:
- bug report? no
- feature request? yes

### Description
Just wondering if you'd be willing to add this to the top of the `shutdownHandler()` function:

```
		// allows defining an alternate shutdown handler
		// set TRACYSDHANLDER to false before calling your alternate handlder
		if (defined('TRACYSDHANDLER') && TRACYSDHANDLER === false) {
			return false;
		}
```

As the comments note, this allows you to specify your own function for `register_shutdown_function` by doing:

```
define('TRACYSDHANDLER', false);
register_shutdown_function('mySpecialShutdownHandler');
```

Maybe you have an idea for a different way to do this, but this works well in my tests.

Thanks for considering.

adrianbj:
Actually a cleaner option is probably this:

```
	/** @var bool whether to disable the shutdown handler in development mode */
	public static $disableShutdownHandler = false;
```

```
	public static function shutdownHandler()
	{
		if (self::$disableShutdownHandler) {
			return false;
		}
```

Then you can simply set:
```
Debugger::$disableShutdownHandler = true;
```
before you want to register your own shutdown handler. 

If you're happy adding this, let me know if you'd like a PR.

dg:
What's the purpose? Shutdown handler is required to catch fatal error and render bar, which can be disabled using $showBar.

adrianbj:
Hi @dg - `$showBar = false` doesn't work in this situation. I have a PHP console panel that runs code via an ajax request. I want this request to be able to update the AJAX bar, but I don't want fatal errors to trigger the bluescreen - I want to handle them myself for this panel. 

If it would help you understand I could post a screencast showing the difference in user experience with and without the ability to use my own shutdown handler.

dg:
You can handle fatal errors by own this way:

```php
register_shutdown_function(function () {
	// handle error 
	error_clear_last();
});

Debugger::enable();
```

adrianbj:
`error_clear_last()` is php 7 only.

dg:
You can "clear" last error in PHP 5 using something like `@trigger_error('', E_USER_NOTICE);`.

I don't like `$disableShutdownHandler` because it reveals internal implementation, as a result, it turns off the Bar etc. So it would be more appropriate disable only fatal-error handler. 

I think it is such an extraordinary requirement that it is better to use the above mentioned "workaround" than directly it incorporate to Tracy.

[...]

PR #334, opened 2018-11-07
Debugger::$disableHandlers, Pjax support

kakserpom:
- bug fix / new feature? feature
- BC break? no

1) $disableHandlers allows to disable register_shutdown_handlers() and set_exception_handler() and use your own handlers.

2) Pjax support (https://github.com/defunkt/jquery-pjax). We treat Pjax requests not like regular Ajax requests.


milo:
ad 1) When you disable Tracy's handlers, what is the purpose of Tracy then? You don't need it enable it after all.

ad 2) It is non systematic solution. You can detect `HTTP_X_PJAX` header in bootstrap and disable Tracy or its bar.

adrianbj:
The concept/need behind this seems somewhat similar to this: https://github.com/nette/tracy/issues/279

I still have to modify the Tracy core on each update to include those changes. I think it wouldn't bet great to have some flexibility to override various handlers.

milo:
> I have my own handlers, but I need Tracy enabled. My own handlers can call Debugger::exceptionHandler($exception, false). Btw, Debugger::exceptionHandler has two arguments and the callback of set_exception_handler() has only one.

You can overload exception handler after Tracy is enabled.
```php
Tracy\Debugger::enable(...);
$prev = null;
$prev = set_exception_handler(function (...) use (&$prev) {
    ...
    if ($prev) $prev(...func_get_args());
});
```

and the same for error handler:
```php
set_error_handler(function (...) {
   ...
   return false;  # Call previous one
});
```
I just want to say, that PHP has already API for it.

> I don't want to disable Tracy, I want to get HTML output for Pjax requests, but they are treated like regular Ajax requests by default.

So why the requests looks like AJAX when they are not? Isn't Pjax library issue? You can always modify `$_SERVER` content to mimic the situation.

PR #335, opened 2018-11-08
Append <div id=tracy-debug> into BODY if possible

EvilKraft:
- bug fix   
- BC break: yes

This fix appends tracy bar into tag BODY if it exists.
Otherwise into documentElement


JanTvrdik:
why?

EvilKraft:
Because of HTML validation. DIV should not be outside BODY

dg:
Thanks

Issue #199, opened 2016-06-20
Only one AJAX request displayed


patrickkusebauch:
Hello,
I upgraded to the latest Tracy version all excited for the new AJAX call functionality.

However it seems from testing in my application (and the example code in source), that only one bar is displayed at the time with the latest AJAX call.

It would be useful if all AJAX calls in the bar lifecycle (one full page reload) could be displayed. However I can understand that this might not be feasible for everybody. I imagine 2 camps:
1. AJAX request are used sporadically to load "time expensive" resources on the site. It might even be several of those calls are sent in parallel. There the need for "bar AJAX history" would be invaluable.
2. AJAX request are used for traversing the website. If possible traversing is not done by full page reload but by redrawing the outdated content (like Nittro). There "bar AJAX history" would unimaginably clutter the bar panel. 

Right now, the implementation favors the second camp. I am interested whether the support for the first camp is in mind or maybe that I just used the bar incorrectly.

dg:
There is important that the man when see 10 ajax bars know which belongs to what request. Which means invent some improved interface.

patrickkusebauch:
I can imagine that the AJAX request URL will be shown (and maybe some additional info) while hovering over the word "AJAX".

radmax:
Request URL is in router panel, why not just add more lines, like with redirects. Max number of AJAX bars should be configurable with some good default value (~3). Bars could also be closed or collapsed.

patrickkusebauch:
I am still open to actually implementing something if we can agree upon how it should look like.

dg:
closed by #341

PR #336, opened 2018-11-09
Show all XHR requests in bar (#199)

josefzajac:
- Show all XHR requests in Bar #199 
- No BC break
- No change in library usage

Preview:
preview
trash icon on end of row deletes all ajax-rows from top up to clicked one. (Icon is not self-explaining, I accept any idea;) )


mabar:
I think there should be an option to limit number of rows.
Also should be possible to call a reset as some sites uses only ajax request and I would expect that bar is cleaned after switching to another page (presenter or view switch)

Anyway, I like the idea 👍

src/Tracy/Bar.php
```diff
@@ -203,10 +204,15 @@ public function dispatchAssets()
 
 		if ($this->useSession && Helpers::isAjax()) {
 			header('X-Tracy-Ajax: 1'); // session must be already locked
+			$seq = mt_rand(1000, 10000000);
```
hranicka: Maybe I'd add timestamp to that sequence to avoid possible collisions. But probability of collision is really really low even with current code, right?

josefzajac: That or increment stored in session will guarantee unique

josefzajac: Start timestamp will collide easily for requests, uniqid brings issues with html#id selectors.

josefzajac:
@dg Hi David, how can I assign ticket to? :)

dakorpar:
What do U think about storing all requests in temp folder and adding just icon in tracy bar which shows then list of all requests stored and when clicking on row you get that bar bellow, while by default just showing last one like now?

josefzajac:
> What do U think about storing all requests in temp folder and adding just icon in tracy bar which shows then list of all requests stored and when clicking on row you get that bar bellow, while by default just showing last one like now?

Hi, sorry, doesn't fit primal needs, eq. I have sequence of xhr requests I want to debug on quick look.
File storage raises complexity in maintenance of this storage (eq. clean up). 
Feel free to fork and ask me for help anytime then ;)

src/Tracy/Bar.php
DanielRuf: Why not use `random_int()`?

josefzajac: I don't think we need crypto-safe number here

DanielRuf: > to avoid possible collisions

Well, random_int() is exactly for this. Other seeding solutions are not collision free.

josefzajac: Ok, implemented, thx

adrianbj:
Really looking forward to seeing this - thanks for the hard work @josefzajac

dg:
I don't know how to collaborate on existing PR, so I created another one #341

DanielRuf:
> I don't know how to collaborate on existing PR, so I created another one #341

https://help.github.com/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork/

And there are also different remotes and aliases and remotes with `pull/...`.
Also see https://gist.github.com/colinsurprenant/9b081958b50cfecc210c

dg:
@DanielRuf thanks!

dg:
Thanks @josefzajac, closed by #341

PR #337, opened 2018-12-11
Recount hiding of labels when resized

dakur:
- new feature
- BC break? no

Calls `Bar.autoHideLabels()` method on every resize of viewport. This is useful when testing web responsivness - you resize the viewport all the time but Tracy does not react with showing/hiding labels when there is (not) enough of space.

Even better would be media query driven showing/hiding.


dg:
Thanks

PR #340, opened 2019-02-15
Dumper: do not quote dumped keys

milo:
Is there any reason for quoting dumped array keys/object property names? For example:
```
array (2)
    Accept => "application/json" (16)
    "User-Agent" => "milo/http-client-v5.2" (21)
```

Empty string is preserver quoted.


dg:
It should preserve quotes in keys like `"hello   "` too.

milo:
Something like this?

dg:
I made it a little easier.

milo:
Great. Thank you!

PR #357, opened 2019-03-12
Nette bridge – fix Container compilation, remove redundant argument

jakubboucek:
- bug fix
- BC break? no
- doc PR: –

Fixes `Nette\InvalidArgumentException` when compile container with `TracyExtension` bridge.

```php
/** @var PhpGenerator\Method $initialize */
$initialize->addBody('$this->getService(?)->start();', [$name, $name]);
//                                     ^^^                   ^^^^^^^
//                                 placeholder         redundant argument
```

Same Bug in branches:
- `master`
- `v2.6`

Bug born at 1f4271623bfabcb4c887dc7f5b61f64b0438a255.


dg:
Thanks, fixed

Issue #356, opened 2019-03-11
Tracy always starts a session


jkufner:
Version: 2.4.0+, introduced by 5c46e989a61fd2594db6eda0f0da4f904ccad4c7

### Bug Description

Tracy always starts a session because https://github.com/nette/tracy/commit/5c46e989a61fd2594db6eda0f0da4f904ccad4c7#diff-be100a67d872789dffa3bfba6144122dR89

### Steps To Reproduce
Don't start a session, enable Tracy.

### Expected Behavior
Session does not start.

### Possible Solution

Remove the start() call from DI container and don't be lazy to check if there is a started session. Such a code don't belong to DI container anyway.

See https://github.com/nette/tracy/commit/5c46e989a61fd2594db6eda0f0da4f904ccad4c7#diff-be100a67d872789dffa3bfba6144122dR89

dg:
Can you send PR?

jkufner:
Maybe. It depends on how deep the session goes. I will give it a try.

milo:
@jkufner It starts in debug mode only, doesn't? https://github.com/nette/tracy/blob/92b97ca42bde7bf40cdf9fcc090ee20ff4951a11/src/Bridges/Nette/TracyExtension.php#L106-L123

jkufner:
@milo Yes, the following line causes the issue:

    $initialize->addBody('$this->getService(?)->start();', [$name]);

milo:
@jkufner Sorry, my point is: Is it a problem to always start session in debug mode? Tracy "always" needs. Usually it is a session for developer on local machine.

jkufner:
@milo If there are no errors to report, then there is no reason to have a session stored and a cookie sent to a client. Tracy <2.4 behaves this way.

Always starting a session prevents us from implementing a state-less API and pollutes the server with useless sessions as every request creates a new session because non-browser clients ignore cookies.

milo:
@jkufner Your server does not run in a production mode?

jkufner:
@milo Of course not; it is a development server.

forrest79:
Hi, we have a similar problem and we would appriciate Tracy without session. I prepared PR - just a concept.

Issue #154, opened 2016-03-24
Option to set own session handler


kukulich:
Tracy doesn't currently work if project runs on more than one server because of this commit: https://github.com/nette/tracy/commit/e1e204c6bc0cb81ada494161e89fb8ea380dc35b

I suggest to add method `setSession()` so it would be possible to set own session handler with shared storage.

dg:
What means `doesn't work`?

kukulich:
Page is generated by one server but request for tracy content is server by another server (because of proxy). However there's no tracy data on the second server because session storage is not shared among servers.

kukulich:
I just tested it and it looks it still does't work.

dg:
Your PHP session storage is shared among servers? Via shared directory or how?

JanTvrdik:
usually redis / memcache / mysql / postgre

Vrtak-CZ:
@dg via memcache

kukulich:
Thanks, it works better now. However there may still be a problem. It works perfectly on one server but ajax requests don't work on our production with session shared via memcached. I'll try to investigate if it's Tracy problem or problem with our environment.

There's also a little BC break. In previous version Tracy can start session, now Tracy requires already started session.

dg:
The fact that Tracy started session annoyed some people, so I created Tracy's own independent session storage. But this storage was not working on your shared servers. So I removed it. And now user must start session manually to see AJAX requests and redirects. 

Session can be started
1) before `Tracy\Debugger::enable()` 
2) after `Tracy\Debugger::enable()`, but you must call `Tracy\Debugger::dispatch()` after that

kukulich:
Thanks a lot for the explanation! That will help me to find the problem.

PR #362, opened 2019-04-02
Tracy without session - concept of using files as a storage [WIP]

forrest79:
- new feature
- issue #356 
- BC break: no

We have a project, where session is started only for logged users. It's a little bit tricky, that Tracy in devel mode always starts session. This could possible hide some bug on devel environment, that is, in better scenario catch with tests, in worse scenario on production :-)

This is just concept, how to let developer choose between session as storage (default and works the same way as now) or file storage. In configuration, you can enable it with setting directory for data files:

```yaml
tracy:
	fileStorage: %tempDir%
```

If this concept will be approved, I can prepare PR for actual Tracy version.

### DIFF
--- a/src/Bridges/Nette/TracyExtension.php
+++ b/src/Bridges/Nette/TracyExtension.php
@@ -21,6 +21,7 @@ class TracyExtension extends Nette\DI\CompilerExtension
 		'fromEmail' => null,
 		'logSeverity' => null,
 		'editor' => null,
+		'fileStorage' => null,
 		'browser' => null,
 		'errorTemplate' => null,
 		'strictMode' => null,
@@ -74,7 +75,7 @@ class TracyExtension extends Nette\DI\CompilerExtension
 		$class = method_exists('Nette\DI\Helpers', 'filterArguments') ? 'Nette\DI\Helpers' : 'Nette\DI\Compiler';
 
 		$options = $this->config;
-		unset($options['bar'], $options['blueScreen'], $options['netteMailer']);
+		unset($options['bar'], $options['blueScreen'], $options['netteMailer'], $options['fileStorage']);
 		if (isset($options['logSeverity'])) {
 			$res = 0;
 			foreach ((array) $options['logSeverity'] as $level) {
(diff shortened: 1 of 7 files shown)


dg:
Thanks a lot for PR.

I need to understand why you're doing this. From my point of view: 

- sessions is a storage on a disk whose identifier is transferred in a cookie.
- your FileStorage is a storage on a disk whose identifier is transferred in a cookie

I simply don't understand why to reimplement session by own class. 

I used to do something like that a long time ago https://github.com/nette/tracy/commit/e1e204c6bc0cb81ada494161e89fb8ea380dc35b#diff-20100846c73ff8ebc44f72ccbbf13c22, but now I don't remember why :-) and there was issue about it #154

forrest79:
I tried to show our reasons in description, maybe not enough clear :-) Imagine web, that in some cases use PHP session (in our case for logged users and some other minor use cases...) but for most requests (~80%), you're not starting session and web needs to cover both variants - session could be running or not. And when you're on your local development machine with Tracy, PHP session is always running, so you can't simply simulate environment without session and is very easy to create a bug. With this PR, you can choose to use "Tracy session" and PHP session is working in the same way as on production web.

dg:
Ok I get it.

forrest79:
I think, that maybe there is no need to have `FileStorage` in Tracy repository, just ability to use your own storage could be also suitable.

ikeblaster:
@dg I have probably found another reason to implement this.

It's actually quite common situation - when you have request with redirect, then the panel is saved into session. Together with memcached it can easily lead to hitting the session size limit. In our case the result is completely deleted session (without any error message visible) and user being redirected to a login page. This was really weird at first and I wonder how nobody has ever met that before.

On production it's obviously not a problem, there is no tracy, we have this on our dev site. So far I have just switched to `files` session handler, but I would rather see better solution. It's desirable to have dev environment as close as possible to production site. I was also thinking about custom session handler, which would split the session between files (just `$_SESSION['_tracy']['redirect']`) and memcached (the rest).

Or is there something I am overlooking?

forrest79:
Hi, I just want to ask, if you have though about this? We're using our fork with this update several months without any problem and right now, we're finishing update to Nette 3, so we will use this also with Tracy 2.7. Thanks for some info.

dg:
I'm going to put it in version 2.8 or the next one.

forrest79:
Great to hear that, thanks 👍

PR #364, opened 2019-04-09
Fix bar display in case that user's css rewrites values in body > div

Rixafy:
- bug fix
- BC break? no

In case that css on page manipulates with `body > div {}` (for instance, width/height), tracy bar would be affected, so bar wrapper now contains `all: initial` in element style.


dg:
fixed

PR #366, opened 2019-04-20
Add support for escaping (for example in json).

janbarasek:
- bug fix
- BC break? yes

Original regex does not support json syntax with escaping in message highliter.

Example:

Snímek obrazovky 2019-04-20 v 12 11 07

New pattern support quotation escaping:

Snímek obrazovky 2019-04-20 v 12 20 28


janbarasek:
@dg

dg:
Thanks

PR #368, opened 2019-05-23
PsrToTracyLoggerAdapter: changed the log message

petaak:
- new feature
- BC break? no

Some loggers (e.g https://github.com/bzikarsky/gelf-php) may require a non-empty log message. Adding the exception class, code and file to the log message solves this problem.


dg:
Thanks

Issue #272, opened 2017-10-26
Catching memory exhaustion


elieux:
- bug report? maybe
- feature request? no
- version: 2.4.8

### Steps to Reproduce
```php
\Tracy\Debugger::enable(\Tracy\Debugger::DEVELOPMENT);
while (true) $a[] = str_repeat('x', 1024);
```

Shows the default PHP error output:
```
Fatal error: Allowed memory size of [...] bytes exhausted (tried to allocate 1025 bytes) in [...] on line [...]

Call Stack: [...]
```

Even worse, it shows nothing with `ini_set('display_errors', false)`.

### Description
The [code that's supposed to make Tracy work even in case of reaching the memory limit](https://github.com/nette/tracy/commit/165a09f46f5d9df761b135b01160ac2fa0e4efd7) doesn't seem to help when the limit is reached by a small allocation.

Is there anything we can do to fix this and let Tracy show the SoD even in this case?

elieux:
#369 made an effort to solve this, but the issue persists.  The problem here is architectural, IMHO it can't be solved by increasing a number.  I tried the with reserved memory of 500'000 B as per the pull request, and it made the red screen appear for the above testcase.  Splitting the allocations in half made Tracy fail again though:

```
\Tracy\Debugger::enable(\Tracy\Debugger::DEVELOPMENT);
while (true) $a[] = str_repeat('x', 512);
```

smuuf:
I think the problem with this is that **any solution based on the amount of pre-allocated memory might never be universally good enough**.

_It's not going to always work_ because different apps - ultimately - will be needing very various amounts of memory to display the Tracy's blue(red)creen in its entirety; apps with small amount of variables (and other stuff) in them will need considerably less memory to render the bluescreen than huge apps with thousands of variables and instantiated classes.

One fixed number will not going to fit all. IMHO it's more about finding the good balance between:
1. ***DO pre-allocate*** a very large amount of memory to match the needs of as many apps/usages as possible.
2. ***DON'T pre-allocate*** such very large amount of memory, because **most of the times the memory is allocated completely unnecessarily** - it matters only when the memory exhaustion finally happens - which it _most_ of the time does not.

The original value before my PR was 30 kB. My PR increased it to 500 kB. It might be a good thing to increase it further to, say, 1 MB of reserved memory. But this increasing might go on forever.

[...]

PR #369, opened 2019-05-25
Increase size of memory reserve for Debugger

smuuf:
The value of 30000 bytes was an incorrect representation of the previous "scientific notation value" of "3e5". I can only assume the notation was changed to increase readability, but the change was done improperly and introduced a regression.

Also the number of bytes reserved for the error renderer was increased to 500 kB, since 300 kBs wasn't enough for the whole error page to render correctly.

- bug fix for issue: *Catching memory exhaustion #272*
- BC break? *Shouldn't be...*

## Before
image
## After
image

### DIFF
--- a/src/Tracy/Debugger/Debugger.php
+++ b/src/Tracy/Debugger/Debugger.php
@@ -152,7 +152,7 @@ class Debugger
 			self::$productionMode = is_bool($mode) ? $mode : !self::detectDebugMode($mode);
 		}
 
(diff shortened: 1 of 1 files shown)


dg:
Thanks

elieux:
Doesn't actually fix #272, just so you know. :)

dg:
@elieux increase Debugger::$reservedMemorySize

PR #372, opened 2019-06-24
Path mapping before using of path.

matlad:
File path mapping shut be done before using of path.  
For actions `fix` and `create` was mapping done after realization of action.

workaround: set mapping in tracy config.

- bug fix 
- BC break? no
- doc PR: not necessary


dg:
Thanks

PR #376, opened 2019-07-12
new cdump() shortcut

iiic:
new cdump() shortcut for debug output into browser's Console using native Debugger::fireLog()

"C" like Console and cdump() is good continuity of bdump()


dg:
Thanks for the contribution, but I don't want to add this shortcut. The existing bdump() covers the browser-side debugging use case, and adding cdump() would expand the API surface without enough benefit. Closing.

PR #377, opened 2019-08-10
InfoPanel: Add information about localhost IP.

janbarasek:
- new feature
- BC break? no

Sometimes beginners have a problem understanding that `127.0.0.1` or `::1` is localhost.

I think Tracy should explain it in brackets:

Snímek obrazovky 2019-08-10 v 22 04 22

Thanks.


dg:
Thanks

PR #382, opened 2019-08-29
Bridge: Remove redundant <pre> and <div>.

janbarasek:
- bug fix
- BC break? no

In case of Latte error in template render multiple of border. Now fixed.

Old render style:

Snímek obrazovky 2019-08-29 v 15 59 59

New style:

Snímek obrazovky 2019-08-29 v 16 02 51

Thanks.


dg:
Thanks

PR #383, opened 2019-09-11
Message max length

janbarasek:
- new feature
- BC break? yes

In case of extremely length error message BlueScreen log full message. I think it should be truncated for example to `32768 characters`.

Example code: `throw new \Exception(str_repeat('abcd ', 10000));`.

Snímek obrazovky 2019-09-11 v 12 50 23

### DIFF
--- a/src/Tracy/Helpers.php
+++ b/src/Tracy/Helpers.php
@@ -291,7 +291,7 @@ class Helpers
 	/** @internal */
 	public static function isAjax(): bool
 	{
-		return isset($_SERVER['HTTP_X_TRACY_AJAX']) && preg_match('#^\w{10,15}\z#', $_SERVER['HTTP_X_TRACY_AJAX']);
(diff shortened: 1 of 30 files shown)


src/Tracy/Helpers.php
```diff
@@ -157,6 +157,43 @@ public static function errorTypeToString(int $type): string
 	}
 
 
+	/**
+	 * Truncates UTF-8 string to maximal length.
+	 */
+	public static function truncate(string $s, int $maxLen = 32768, string $append = "\u{2026}"): string
```
JanTvrdik: This is too complex for Tracy. You should extract the truncate code from Dumper

JanTvrdik: https://github.com/nette/tracy/blob/44349212f8856cf1cacb0ab2b62c2b30e4be6f93/src/Tracy/Dumper/Dumper.php#L519-L531

dg: I added method Dumper::truncateString()

janbarasek: @dg Thank you and replaced `Helpers::truncate` to `Dumper::encodeString`.

dg:
thanks

PR #384, opened 2019-09-12
Bridge: Template line can be null.

janbarasek:
- bug fix
- BC break? no

In case of template error line is null Tracy can not render filename:

Snímek obrazovky 2019-09-12 v 17 06 13

Fixed to:

Snímek obrazovky 2019-09-12 v 17 10 22

### DIFF
--- a/src/Bridges/Nette/Bridge.php
+++ b/src/Bridges/Nette/Bridge.php
@@ -52,7 +52,7 @@ class Bridge
 			$lines = file($file);
 			if (preg_match('#// source: (\S+\.latte)#', $lines[1], $m) && @is_file($m[1])) { // @ - may trigger error
 				$templateFile = $m[1];
-				$templateLine = preg_match('#/\* line (\d+) \*/#', $lines[$e->getLine() - 1], $m) ? (int) $m[1] : null;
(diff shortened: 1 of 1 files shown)


src/Bridges/Nette/Bridge.php
```diff
@@ -52,7 +52,7 @@ public static function renderLatteError(?\Throwable $e): ?array
 			$lines = file($file);
 			if (preg_match('#// source: (\S+\.latte)#', $lines[1], $m) && @is_file($m[1])) { // @ - may trigger error
 				$templateFile = $m[1];
-				$templateLine = preg_match('#/\* line (\d+) \*/#', $lines[$e->getLine() - 1], $m) ? (int) $m[1] : null;
+				$templateLine = preg_match('#/\* line (\d+) \*/#', $lines[$e->getLine() - 1] ?? '', $m) ? (int) $m[1] : null;
```
dg: Imho this is more readable `$e->getLine() && preg_match('#/\* line (\d+) \*/#', $lines[$e->getLine() - 1], $m) ? (int) $m[1] : null;`

janbarasek: Fixed, thanks.

dg:
Thanks

PR #391, opened 2019-10-12
readme: info about error_reporting(E_ALL) is used

holtkamp:
Which I found out after noticing that "somehow" the error reporting level of my application changed. Traced it down to Tracy 😄


dg:
Thanks

PR #392, opened 2019-11-01
BlueScreen HTTP Request: Add support for rendering PHP Input

janbarasek:
- new feature
- BC break? yes

If request was called as ajax json payload with some data but without `'Content-Type': 'application/x-www-form-urlencoded'` header, Tracy can not dump sent values.

This change add automatic detection of `PHP Input` and render to `HTTP request` section.

Example:

Snímek obrazovky 2019-11-01 v 12 12 52

If entered data was short json, Tracy will detect it automatically and dump as array for bette readability.

More information (Czech language): https://php.baraja.cz/ajax-post

Thanks.

### DIFF
--- a/src/Tracy/BlueScreen/assets/content.phtml
+++ b/src/Tracy/BlueScreen/assets/content.phtml
@@ -309,6 +309,18 @@ $code = $exception->getCode() ? ' #' . $exception->getCode() : '';
 			</table>
 			</div>
 
+			<?php
+			if (($phpInput = (string) file_get_contents('php://input')) !== ''):
+			?>
+			<h3>PHP Input</h3>
+			<pre class="code"><?= Helpers::escapeHtml($phpInput) ?></pre>
+			<?php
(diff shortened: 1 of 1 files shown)


lukaspijak:
Hi it's nice feature but this "JSON detection" is very dumb. I think it would be better to detect JSON using content type (application/json).

JanTvrdik:
You cannot simple use `file_get_contents('php://input')`, because the output may be too large.

janbarasek:
Thanks to everybody.

@lukaspijak Added detection by HTTP header `Content-Type` which must starts with `application/json`. For example: `application/json;charset=UTF-8`

@JanTvrdik Added limitation for 10000 input characters.

KminekMatej:
+1

dg:
Thanks! This is already implemented — when `$_POST` is empty on a POST request, Tracy reads `php://input` and shows it as a POST preview. Closing.

PR #399, opened 2020-01-08
Debugger::log(), parameter $level can be null

janbarasek:
- new feature?
- BC break? yes

I think second parameter of `Debugger::log()` can be null.

In case of simple call `Debugger::log($e);` in user code Tracy should log it as `Exception` level and in case of `Debugger::log('Message');` as default `Info`.

It can be detected by `nullable` default value.


dg:
I see no compelling reason to do a serious BC break…

janbarasek:
Ok.

PR #400, opened 2020-01-12
Content: Add link as new window

janbarasek:
- new feature
- BC break? no

Added icon for open link to new window:

Snímek obrazovky 2020-01-12 v 22 35 44

### DIFF
--- a/src/Tracy/BlueScreen/assets/content.phtml
+++ b/src/Tracy/BlueScreen/assets/content.phtml
@@ -298,12 +298,7 @@ $code = $exception->getCode() ? ' #' . $exception->getCode() : '';
 
 		<div class="tracy-collapsed inner">
 			<h3>URL</h3>
-			<p>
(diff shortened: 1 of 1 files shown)


dg:
I think it's better to open it in a new window every time. The icon seems to belong to the link.

janbarasek:
@dg It's make sense. Fixed.

src/Tracy/BlueScreen/assets/content.phtml
```diff
@@ -298,12 +298,7 @@ $code = $exception->getCode() ? ' #' . $exception->getCode() : '';
 
 		<div class="tracy-collapsed inner">
 			<h3>URL</h3>
-			<p>
```
dg: Please can you add `rel="noreferrer noopener"` after `target="_blank"`?

janbarasek: Added.

dg:
Thanks

PR #401, opened 2020-01-29
Info panel: Add server name by IP

janbarasek:
- new feature
- BC break? yes

Very often I need to find out what production server the site is running on. This feature should be very useful.


dg:
Thanks

PR #402, opened 2020-01-29
Info panel: When server IP is localhost add warning.

janbarasek:
- new feature
- BC break? yes

Sometimes server IP can be `127.0.0.1`, because server admin make mistake in configuration.

For instance:

Snímek obrazovky 2020-01-29 v 16 27 49

This feature add warning message like Tracy in argument block.

### DIFF
--- a/src/Tracy/Bar/panels/info.panel.phtml
+++ b/src/Tracy/Bar/panels/info.panel.phtml
@@ -32,6 +32,13 @@ $ipFormatter = static function (?string $ip): ?string {
 	return $ip;
 };
 
+$serverNameFormatter = static function(string $userIp, string $serverIp): string {
+	if ($serverIp !== '127.0.0.1' && $serverIp !== '::1') {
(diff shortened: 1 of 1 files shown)


midlan:
I am voting against this PR
@janbarasek I did not catch what you are actually trying to achive with the `` but it is wrong in many ways, it matches invalid addresses:

```
127.0.0.01
127.0.0.001
127.0.0.999
...
```

Aside that problem I guess you are trying identify local IP addresses. Your regex is matching only addresses in `127.0.0.0/24` network, but the netwrok of local addreses is much wider, in fact it is `127.0.0.0/8`. And the last note: DNS name `localhost` is always only the `127.0.0.1` address.

I like you moved the logic inside the function $serverNameFormatter, but the newly added logic is broken.

janbarasek:
@midlan Yes, I think this PR can be closed, because I solve it by #406.

PR #403, opened 2020-02-05
Logger: Remove shut-up marker on date()

jakubboucek:
PHP 7+ no more thrown warning when timezone not set.

- new feature
- BC break? no
- doc PR: not needed

## Description
- Nette Tracy is required PHP 7 +
- PHP 7 is no mote throw warning on using `date`-like functions without defined current TimeZone
- Silent error marker on `@date()` function is no more needed.

## Sources
- https://hakre.wordpress.com/2016/01/03/history-of-the-php-date-timezone-settings-warning/
- https://github.com/php/php-src/pull/1029

Thanks for this great tool ❤️


dg:
Thanks!

PR #404, opened 2020-02-12
Bridge: When template line is null, ignore whole Template section.

janbarasek:
- bug fix
- BC break? yes

I think in case when I can not detect position of error, template section should be ignored.

Current behavior:

Snímek obrazovky 2020-02-12 v 16 39 02

New behavior:

Snímek obrazovky 2020-02-12 v 16 38 40

Thanks.


dg:
I find it useful to know which template has an error. In addition, I can easily open it in editor.

PR #406, opened 2020-02-28
Info panel: Better detection server IP and server name.

janbarasek:
- bug fix
- BC break? no

I think this is more better way how to detect real server IP and real server name. Supports simple server architecture and Ngnix as reverse proxy.

Explanation: https://php.baraja.cz/ip-adresa (czech language).

Snímek obrazovky 2020-02-28 v 10 38 35

Thanks.


dg:
Thanks

PR #411, opened 2020-04-07
Debugger::enable(): More helpful comment

janbarasek:
- bug fix
- BC break? no

I think this comment is more helpful, because first parameter with boolean is often source of misunderstand.

Thanks.


dg:
Wouldn't it be better to say that one of the constants `Debugger::PRODUCTION`, `DEVELOPMENT`, `DETECT` should be used?

janbarasek:
@dg Replaced. Thanks.

dg:
Thanks

PR #420, opened 2020-05-15
bar.js: fetch checks Response

HamAndRock:
In project I am currently working on, I also wrap `fetch()` method and return the result as `json` object, but tracy doesn't check if the response is actually a Response object.

So it thinks my json object is `Response` object and tries to access `.headers` property, that may not exist and then run function `has()` on it.

- bug fix
- BC break? no


dg:
Thanks

PR #432, opened 2020-09-10
Only predefined log levels should be valid

mabar:
- bug fix
- BC break? only for static analysis

Default logger works with custom log levels, but the PSR adapters converts custom levels to error. This behavior could be unexpected and switch from default to PSR leads to break of custom levels logic.

Imho the PSR adapter logic makes sense for BC and makes switch between loggers easier but the PSR-3 specifies for  `Psr\Log\LoggerInterface::log()` following:
> Calling this method with a level not defined by this specification MUST throw a Psr\Log\InvalidArgumentException if the implementation does not know about the level.

This PR hints phpstan using codebases to use only defined levels and prevent the error.

I would also like to deprecate custom log levels completely and throw an exception, but it's probably not a good idea to create such BC break in logger


dg:
Custom log levels are widely used, basically, it's not about levels, but about categorizing logs.

mabar:
It's kinda unfortunate way how to categorize them as severity and category of the logged incident are two different things, but I understand. Feel free to close.

dg:
I'm not saying it's a good way, but if custom levels were to be deprecated, an alternate path would have to be devised first.

JanTvrdik:
> Default logger works with custom log levels, but the PSR adapters converts custom levels to error. This behavior could be unexpected and switch from default to PSR leads to break of custom levels logic.

We could make the [level map](https://github.com/nette/tracy/blob/v2.7.5/src/Bridges/Psr/PsrToTracyLoggerAdapter.php#L21-L29) extendable (e.g. by switching from private const to public property), so you can map you application-specific custom level to whatever is appropriate in PSR. Would that solve your issue?

mabar:
With Monolog I would implement it probably like that
`$logger->withName($category)->log($severity, $message)` which does not seem suitable for Tracy
or `$logger->log($severity, $message, ['category' => $category]);` which is bit too much verbose

`Debugger::log($message, $priority, $category)` could work, but it's not much future-proof

Probably even optional check by phpstan/phpstan-nette would be enough

mabar:
> We could make the level map extendable (e.g. by switching from private const to public property), so you can map you application-specific custom level to whatever is appropriate in PSR. Would that solve your issue?

Not really. I am looking for a way to get rid of custom levels. Mapping of anything to one level breaks logic which were originally behind these custom levels.

But I maybe get the right idea thanks to you. Custom levels which can't be mapped without an error can add context to logged messages. This way, a PSR logger can at least handle it internally, because information about original level is not lost in the process.

Issue #430, opened 2020-09-07
BlueScreen exposes sensitive ENV variables


dakujem:
Version: 2.6.x, 2.7.x

### Bug Description
`BlueScreen` reveals sensitive server variables without offering option to turn this feature off or any kind of filtering option.
Furthermore, half of the class' code is private, so no easy customisation .... yet again.

### Steps To Reproduce
Cause an exception to be rendered using `BlueScreen`, observe all server vars dumped at the bottom of the rendered screen.

### Expected Behavior
Filtering of sensitive data enabled. Sensitive data not shown in plaintext.

### Possible Solution
Refactor. Introduce options. Enable extensibility. Avoid private members and methods when no replacement is possible using composition.

### Why Care
In cloud environments, it is common and/or necessary to store configuration options (like database connections, service keys, third party api tokens, etc. ) in environment variables. The described issue poses a security risk. Read more here: [12 factor app](https://12factor.net/config).

dakujem:
This is quite possibly a security feature-issue. Why no response?

dakujem:
I looked into this deeper and found out that the variables in the Environment section actually _are_ filtered, except that no variable is matched (see the issue #431 for details). I prepared a PR #438 to patch this issue.

PR #438, opened 2020-09-22
Allow to hide the Environment section by configuration

dakujem:
This is a new feature that allows us to omit the Environent section from the output.

```php
// hide the environment section (shown by default)
Tracy\Debugger::getBlueScreen()->showEnvironment = false;
```

This refers to #430 issue.


dg:
Thanks

PR #439, opened 2020-09-22
Scrubber mechanism

### LINKED ISSUE #431: Improve scrubbing/filtering capabilities of BlueScreen
### Problem
BlueScreen currently only filters fields that match certain strings, like `password`, or  `pwd`.

```php
		$keysToHide = array_flip(array_map('strtolower', $this->keysToHide));

		return function ($v, $k = null) use ($keysToHide): string {
			if (is_string($k) && isset($keysToHide[strtolower($k)])) {
				$v = Dumper::HIDDEN_VALUE;
			}
```

Unfortunatelly, it does not even cover a simple use case where a password is typed and re-typed for confirmation.
```json
{
    "username":"foobar",
    "password":"*******",
    "password_check":"my-Secret_p4sswoad"
}
```
### Solution:
1. In addition to keys-to-hide list, introduce a callback option so that devs can come up with their own scrubbing strategy (regexp for example).
2. Use `strpos` instead of `isset` to include strings like `password_check` by default.
3. Include `secret` and `dsn` in the `BlueScreen::$keysToHide` list by default.

---
I'm willing to provide a PR if it gets constructive attention quickly.

dakujem:
This feature adds possibility to define/configure a custom "scrubber" in a flexible way.
A scrubber allows for custom variable filtering to prevent sensitive data leak.
It expands the currently built-in "keys to hide" scrubbing mechanism, and should IMHO replace it by next major release.

This allows for (among others):
- scrubbing based on value and/or keys (current scrubbing only works with keys)
- scrubbing based on regexp matching (current scubbing only matches exact match, case-insensitive)

```php
Tracy\Debugger::getBlueScreen()->scrubber = function(string $key, $val = null): bool {
    return
        # matches `password`, `password_repeat`, `check_password`, `DATABASE_PASSWORD`, ....
        preg_match('#password#i', $key) ||
        # matches DSN defined connections for services etc.
        ( is_string($val) && preg_match('#^https://#i', $val) )
    ;
}
```
... and any logic users want to _easily_ fit into it.

> Backwards compatibile.

Closes #431

### DIFF
--- a/src/Tracy/BlueScreen/BlueScreen.php
+++ b/src/Tracy/BlueScreen/BlueScreen.php
@@ -29,6 +29,9 @@ class BlueScreen
 	/** @var int  */
 	public $maxLength = 150;
 
+	/** @var callable|null  a callable returning TRUE for sensitive data; fn(string $key, mixed $val = null): bool */
+	public $scrubber;
+
 	/** @var string[] */
 	public $keysToHide = ['password', 'passwd', 'pass', 'pwd', 'creditcard', 'credit card', 'cc', 'pin'];
 
@@ -360,9 +363,12 @@ class BlueScreen
 	public function getDumper(): \Closure
 	{
 		$keysToHide = array_flip(array_map('strtolower', $this->keysToHide));
+		$scrubber = $this->scrubber ?? function (string $k) use ($keysToHide): bool {
+			return isset($keysToHide[strtolower($k)]);
+		};
 
-		return function ($v, $k = null) use ($keysToHide): string {
-			if (is_string($k) && isset($keysToHide[strtolower($k)])) {
+		return function ($v, $k = null) use ($scrubber): string {
+			if (is_string($k) && $scrubber($k, $v)) {
 				$v = Dumper::HIDDEN_VALUE;
 			}
 			return Dumper::toHtml($v, [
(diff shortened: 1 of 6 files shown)


src/Tracy/BlueScreen/BlueScreen.php
```diff
@@ -360,17 +363,20 @@ public function isCollapsed(string $file): bool
...
 	{
 		$keysToHide = array_flip(array_map('strtolower', $this->keysToHide));
+		$scrubber = $this->scrubber ?? function (string $k) use ($keysToHide): bool {
+			return isset($keysToHide[strtolower($k)]);
+		};
 
-		return function ($v, $k = null) use ($keysToHide): string {
-			if (is_string($k) && isset($keysToHide[strtolower($k)])) {
+		return function ($v, $k = null) use ($scrubber): string {
+			if (is_string($k) && call_user_func($scrubber, $k, $v)) {
```
JanTvrdik, line 371: `$scrubber($k, $v)`

dakujem: Why? I'm not against, just don't know the reasoning.

dakujem: Okay, this seems to be a bad habit from PHP 5, even thogh it is only about negligible 1e-10 seconds faster, I updated the code.

dakujem:
Added tests that improve code coverage.

dakujem:
Is there any way to push this PR forward? @JanTvrdik 
It is quite critical for a client and we had to fall back to a workaround without tracy for the time being.
Thanks a lot!

dg:
Thanks, nice implementation!

PR #440, opened 2020-10-02
Info panel JIT

MikyWoW:
- new feature
- BC break? no

PHP 8 is supports JIT. Would be nice to show if its enabled.


dg:
Thanks

diegosardina:
```
Error: Tracy:info
DivisionByZeroError: Division by zero in [...]/vendor/tracy/tracy/src/Tracy/Bar/panels/info.panel.phtml:47

```
This should be improved a bit more, because the default configuration and behavior is weird.

Jit is enabled by default (in tracing), but [buffer_size](https://www.php.net/manual/en/opcache.configuration.php#ini.opcache.jit-buffer-size) is 0 (that is its default value) and the zero value disables JIT. Then JIT is not enabled although $opcache['jit']['enabled'] is true.

It should be enough to check for $jit['buffer_size'] > 0

dg:
@diegosardina I tried to fix it

PR #445, opened 2020-10-27
Info panel: Max table width is 300px + better responsivity

janbarasek:
- bug fix
- BC break? no

In case of long string (for example `Server`) old panel should break responsivity like this:

Snímek obrazovky 2020-10-27 v 21 34 07

But new CSS styles should format the table to normal size:

Snímek obrazovky 2020-10-27 v 21 35 03

Thanks.


dg:
Merged, thanks

PR #446, opened 2020-11-03
Helpers::editorLink(): File path can contains matched line

janbarasek:
- new feature
- BC break? no

In the case of a custom Tracy panel with the highlighted line, it can be a very useful feature.

*Be conservative in what you send, be liberal in what you accept.*


dg:
I see it as unnecessary complexity.

PR #450, opened 2020-12-09
[bugfix] Bar: fix clickable link in sortable tables

jakubboucek:
- bug fix
- BC break: no
- Target version: `v2.7` because it's bug

Currently is unable to open `editor:` link in Tracy error panel, because the TracySort handler is catching that click.

image

This PR is adding table header for Error panel to prevent conflict links in content with sorting feature.

image

### Note:
Project contains more sortable tables, only error panel contains anchors and sorting feature together.


dg:
Thanks

PR #452, opened 2020-12-21
Debugger: The value of $maxDepth should be higher in production or CLI mode

janbarasek:
- new feature
- BC break? yes

In production mode, we do not have the option to increment the value of $maxDepth, which can lose important information (such as the internal state of services and databases) when saving objects to the log.

Wow, GitHub now supports videos:

https://user-images.githubusercontent.com/4738758/102800514-45d9ad80-43b4-11eb-8f0b-f4ffb43e5c64.mov

### DIFF
--- a/src/Tracy/Debugger/Debugger.php
+++ b/src/Tracy/Debugger/Debugger.php
@@ -524,17 +524,25 @@ class Debugger
 	 */
 	public static function dump($var, bool $return = false)
 	{
+		$maxDepth = self::$maxDepth;
+		if (self::$productionMode === true || PHP_SAPI === 'cli') {
+			if ($maxDepth < 3) {
+				$maxDepth = 3;
+			} elseif ($maxDepth < 10) {
+				$maxDepth *= 2;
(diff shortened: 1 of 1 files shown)


dg:
In the last version it is higher.

dg:
> In production mode, we do not have the option to increment the value of $maxDepth

Why not? Debugger::$maxDepth works in production mode too, or not?

janbarasek:
@dg Yes, but in production mode, `Debugger::$maxDepth` should be higher for dump more details to a log file.

dg:
But you can set it up extremely easily, and there's no need to add *magic* to Tracy.

janbarasek:
@dg OK I understand. It was an idea for a systemic solution to ensure that structures were dumped in greater depth in production mode.

What about adding the static property `$maxDepthProduction`, which will have a default value about 30% higher than `maxDepth`?

dg:
Simply set $maxDepth to 30 % higher value in production.

PR #464, opened 2021-02-02
open-editor.js: added Visual Studio Code/VS Codium

Tezar:
Extended open-editor.js with Visual Studio Code/VS Codium example.


dg:
Thanks

PR #469, opened 2021-03-01
Debugger::enable(): Fix broken autoload [bug]

janbarasek:
- bug fix
- BC break? no

While installing Nette Sandbox in Docker, I noticed that in some cases autoload may not be available and classes may not load correctly.

The original implementation goes through an array of strings, but they are not used anywhere and only loads the Debugger each time:

Screenshot from 2021-03-01 15-26-48

The original implementation used a non-existent path (logic error):

Screenshot from 2021-03-01 15-26-37

Now I have fixed the whole problem.

Thanks.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -26,12 +26,14 @@ If you are meeting Tracy for the first time, believe me, your life starts to be 
 Documentation can be found on the [website](https://tracy.nette.org).
 
 
-Support Project
----------------
+[Support Tracy](https://github.com/sponsors/dg)
+-----------------------------------------------
 
 Do you like Tracy? Are you looking forward to the new features?
 
-[![Donate](https://files.nette.org/icons/donation-1.svg?)](https://nette.org/make-donation?to=tracy)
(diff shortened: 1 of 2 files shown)


dg:
I don't understand what it solves.

janbarasek:
@dg When a Tracy instance was called during the Composer autoload task (and the autoloader has not yet been built), in some cases it was not possible to load all the files. This MR fixes loading all classes correctly (previously there was an error in the file path).

janbarasek:
Problem is here (because you ignore iterated classes and always load only the Debugger):

Screenshot from 2021-03-08 15-59-59

dg:
What?

https://github.com/nette/tracy/blob/0c13775ea0d3b681ae1d9bea59ed94029a96c09d/src/Tracy/Debugger/Debugger.php#L212

janbarasek:
Oh, I'm so sorry! I debugged Tracy on the legacy project where this error occurs (screenshot comes from that project). :/

dg:
@mesour found out that PHPStorm is causing this.

PR #471, opened 2021-03-04
Bluescreen: Add selection color

janbarasek:
- new feature
- BC break? no

Hello,

on some systems (like Ubuntu) the user may have a red color to selection the text.

This fix sets the default color for the selection to always the same yellow (according to the code background) and the text as black.

Original example:

Screenshot from 2021-03-04 14-59-55

New example:

Screenshot from 2021-03-04 15-05-25

Thanks.


dg:
thanks

PR #473, opened 2021-03-18
Dumper: scrubber checks also non-string keys

vojtech-dobes:
- bug fix
- BC break? no

We are using Tracy support for scrubbing to remove passwords etc. from our Tracy bluescreen files. But we've found that in following part of stacktrace (in `Elasticsearch\Transport` from official Elasticsearch SDK), password remained:

```
connectionParams: array
    'client' => array
        'x-elastic-client-meta' => true
        'curl' => array
            107 => 1
            10005 => 'actualHostHere:actualPasswordHere'
```

It's because `Describer` doesn't apply scrubber on non-string keys. This PR argues that they have to be for sanity checked as well. I've chosen the path of not changing Scrubber signature, so I cast scalar keys to strings and pass empty string for non scalar keys. To be honest I am not sure whether such eventuality is possible (I was thinking about `WeakMap` and stuff like that, but not sure if it would make its way there).


vojtech-dobes:
@dg Is there something I can do to make this acceptable change? I can imagine the issue is that this makes the scrubber signature somewhat less clearly defined?

dg:
Thanks

vojtech-dobes:
Thank you

PR #477, opened 2021-04-05
Update Debugger: apply editor mapping for browser from CLI

thorewi:
- bug fix 
- BC break: no

When opening browser from CLI, editor mappings are not applied.


dg:
Why??

thorewi:
Because the path in docker may be (and probably is) different than on the host and then the browser is not opened.

dg:
Ok, thanks

PR #480, opened 2021-04-21
Logger: also drop 'object' before serialization

vojtech-dobes:
- bug fix
- no BC break

When stack trace contains `'object'` field (which is awesomely supported for many years :) ), `Logger` doesn't get rid of it before serialization when establishing exception file name. This PR removes the `'object'` field in same way as `'args'` is being dropped already.

### DIFF
--- a/src/Tracy/Logger/Logger.php
+++ b/src/Tracy/Logger/Logger.php
@@ -124,7 +124,7 @@ class Logger implements ILogger
 		while ($exception) {
 			$data[] = [
 				get_class($exception), $exception->getMessage(), $exception->getCode(), $exception->getFile(), $exception->getLine(),
-				array_map(function (array $item): array { unset($item['args']); return $item; }, $exception->getTrace()),
+				array_map(function (array $item): array { unset($item['args'], $item['object']); return $item; }, $exception->getTrace()),
 			];
 			$exception = $exception->getPrevious();
 		}


dg:
What you mean by *which is awesomely supported for many years*?

vojtech-dobes:
> Supported by PHP?

Supported by Tracy.

dg:
Look, I don't understand you at all :-)

I've never seen an 'object' field in exception callstack. Maybe you have a PHP extension that puts it there?

vojtech-dobes:
@dg Yes, there is no simple way to get it there, only `debug_backtrace` generates that. I was only referring to the fact that Tracy is aware of possible existence of that field :). And this PR is only about the discovery that when it's truly present (in some non-standard magical way), other part can break (serialization). I would love if this was accepted. I can imagine it could be accepted because it has I believe no impact on code that doesn't get somehow (magically) `object` field populated.

dg:
I'm sorry, but I feel like you're wasting my time.

vojtech-dobes:
@dg that was definitely not my intention, so sorry for that. But can you please tell me, is there some reason why same argument wouldn't apply for existence of https://github.com/nette/tracy/blob/master/src/Tracy/BlueScreen/assets/content.phtml#L108? I was basing my reasoning entirely on existence of this "support".

dg:
Ah, I get it. It is probably some historical relict…

PR #484, opened 2021-05-06
Dumper: added ext-ds support

### LINKED ISSUE #472: ext-ds support
Instead of arrays I much prefer using specialized data structures from ext-ds extension:

- https://github.com/php-ds/ext-ds
- https://www.php.net/manual/en/book.ds.php
- https://pecl.php.net/package/ds

Unfortunately Tracy dumper seems unable to show me the contents of the data structures which often leads to important info to be missing on the exception screen.

If I understand Tracy correctly then this should be possible using [`\Tracy\Dumper::$objectExporters`](https://github.com/nette/tracy/blob/c0f21d61bec95984470eeceee7d3e4496f3dee7f/src/Tracy/Dumper/Dumper.php#L74).

enumag:
Closes #472

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,6 +22,7 @@ jobs:
             - uses: shivammathur/setup-php@v2
               with:
                   php-version: ${{ matrix.php }}
+                  extensions: ds
                   coverage: none
 
             - run: composer install --no-progress --prefer-dist
@@ -41,6 +42,7 @@ jobs:
             - uses: shivammathur/setup-php@v2
               with:
                   php-version: 7.4
+                  extensions: ds
                   coverage: none
 
             - run: composer install --no-progress --prefer-dist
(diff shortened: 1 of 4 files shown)


enumag:
@dg The failure is not caused by my changes.

dg:
Can you add a test?

enumag:
The test would of course require ext-ds to be installed. I'm not sure how to add it on your GitHub Actions setup.

dg:
In this way: https://github.com/nette/latte/blob/master/.github/workflows/tests.yml#L20

(Probably `extensions: ds`)

enumag:
@dg How do I run the tests locally? I tried `composer tester` but it fails with

```
   Error: Interface 'JsonSerializable' not found
```

I have no idea why. I'm using json extension every day and in my code it always exists. So I'm very confused. Can you fix `composer tester` so that it would work?

milo:
`vendor/bin/tester -C tests` should do the job

dg:
ad JsonSerializable: what operating system are you using?

enumag:
@dg Ubuntu but I don't think it's important. The issue is that nette/tester isn't using the system's default php.ini without the `-C` option so it didn't load any extensions. It is very counter-intuitive for someone not experienced with nette/tester.

dg:
The point is, it works for me on Windows and I had no idea there is such a problem. 

This has to be solved with a custom php.ini, I'll fix it so please don't change the composer.json.

PR #486, opened 2021-05-19
Debugger: shows error message in console

mabar:
I just don't think there is a reason to not show error message in console. But I may be wrong, let me know if I missed something.

### DIFF
--- a/src/Tracy/Debugger/Debugger.php
+++ b/src/Tracy/Debugger/Debugger.php
@@ -319,8 +319,12 @@ class Debugger
 					require self::$errorTemplate ?: __DIR__ . '/assets/error.500.phtml';
 				})(empty($e));
 			} elseif (PHP_SAPI === 'cli') {
-				@fwrite(STDERR, 'ERROR: application encountered an error and can not continue. '
-					. (isset($e) ? "Unable to log error.\n" : "Error was logged.\n")); // @ triggers E_NOTICE when strerr is closed since PHP 7.4
+				// @ triggers E_NOTICE when strerr is closed since PHP 7.4
+				@fwrite(STDERR, "ERROR: {$exception->getMessage()}\n"
+					. (isset($e)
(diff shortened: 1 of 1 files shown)


mabar:
Output looks like this:
```
ERROR: Ello there, I am veri broken app.
Error was logged. Enable debug mode to see more info.
```

Build failure is unrelated to the PR, symplify/ecs seems to be broken.

mabar:
I usually enable debug mode, to see stack trace, etc. Writing into log may fail.

What about `To see more, enable debug mode.` and `To see more, enable debug mode or check log.` in case error was logged?

dg:
If it's been logged, there's no need to do anything.

mabar:
> If it's been logged, there's no need to do anything.
Good point, changed.

With log
```
ERROR: Ello there, I am veri broken app.
Check log to see more info.
```

Without log
```
ERROR: Ello there, I am veri broken app.
Unable to log error. You may try enable debug mode to inspect the problem.
```

dg:
Great!

src/Tracy/Debugger/Debugger.php
```diff
@@ -319,8 +319,12 @@ public static function exceptionHandler(\Throwable $exception): void
 					require self::$errorTemplate ?: __DIR__ . '/assets/error.500.phtml';
 				})(empty($e));
 			} elseif (PHP_SAPI === 'cli') {
-				@fwrite(STDERR, 'ERROR: application encountered an error and can not continue. '
-					. (isset($e) ? "Unable to log error.\n" : "Error was logged.\n")); // @ triggers E_NOTICE when strerr is closed since PHP 7.4
+				// @ triggers E_NOTICE when strerr is closed since PHP 7.4
```
Ciki, line 322: there's a typo `strerr` should be `stderr`

dg: serepes

PR #487, opened 2021-05-20
Debugger: Add property $enableDispatch for disable dispatching.

janbarasek:
- bug fix
- BC break? no

In some Tracy integrations outside the Nette framework, we need to ensure that the request is not called twice when downloading assets for Tracy.

Outside of Bootstrap, this behavior is not guaranteed, so there should be an easy way to disable dispatch but also have active debug mode.

Thank you.

### DIFF
--- a/src/Tracy/Debugger/Debugger.php
+++ b/src/Tracy/Debugger/Debugger.php
@@ -40,6 +40,9 @@ class Debugger
 	public static $reservedMemorySize = 500000;
 
 	/** @var bool */
+	public static $enableDispatch = true;
+
+	/** @var bool */
 	private static $enabled = false;
 
 	/** @var string|null reserved memory; also prevents double rendering */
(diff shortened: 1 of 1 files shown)


mabar:
Couldn't it be internal behavior? It does not make sense to call that method more than once a request anyway, method could handle it itself with private variable.

janbarasek:
I may have explained it wrong.

If Tracy registers after another function is called in Bootstrap, the second request to download assets will result in a second call to the action.

For example, if an inappropriately designed router in a hand-written application is first run and calls the logic to insert into the database, the first request will first execute the regular request (and display Tracy), however, immediately afterwards a second request (to download assets) is called, but that will also trigger the insert.

dg:
I certainly don't think the correct solution is to add a new variable, but simply to solve the problem.

PR #488, opened 2021-06-01
Enrich tracy error mail with request method

KminekMatej:
- new feature
- BC break? no
- doc PR: nette/docs# 2B after approval

When using nette/tracy for API project, having request method in tracy error mail is for us highly valuable information. I simply injected this information right after source url. If the method was GET, nothing gets injected (kinda like default method). Method gets injected only for all other methods

### DIFF
--- a/src/Bridges/Nette/MailSender.php
+++ b/src/Bridges/Nette/MailSender.php
@@ -50,7 +50,7 @@ class MailSender
 			$mail->addTo(trim($item));
 		}
 		$mail->setSubject('PHP: An error occurred on the server ' . $host);
-		$mail->setBody(Tracy\Logger::formatMessage($message) . "\n\nsource: " . Tracy\Helpers::getSource());
+		$mail->setBody(Tracy\Logger::formatMessage($message) . "\n\nsource: " . Tracy\Helpers::getSource() . Tracy\Helpers::formatMethod());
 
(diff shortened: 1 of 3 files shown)


dg:
I would prefer a more general solution, where you could redefine the content of the mail yourself.

KminekMatej:
Well, me too :). That means you have it on the roadmap, so I shoudlnt participate?

dg:
No, I don't. Try to come up with a solution that works for you.

janbarasek:
@KminekMatej I think in Tracy should exist interface `EmailRenderer` with default implementation for render e-mail message.

For example:

```php
interface EmailRenderer
{
    public function render(\Throwable $e, ?string $message, string $level): string;
}
```

KminekMatej:
Something like that, yea. I've been thinking about possibility to render email body using user-specified latte template.
User would simply put path to latte file into config.neon (somewhere underneath the tracy section in config). If no latte template would be specified, some default one should be used - or maybe just current solution would be used (simply text, no latte rendering).

This template would have some important variables injecte, like httpRequest, httpResponse, etc.

But I really didnt start to design any implementation as Ive been pretty busy lately, just made some private brainstorming

dg:
It's not such a requested feature, so a callback instead of an interface would probably be enough…

PR #495, opened 2021-06-17
Logger: Add current datetime to e-mail log message.

janbarasek:
- new feature?
- BC break? no

Just adding a datetime to better reveal when the error has occurred.

Sending an email can have a delay, this can reveal the real time the error was caught right in the email.


dg:
Related to #488 

With @KminekMatej, try to prepare solution how to easily change the content of the sent email.

dg:
Closing due to inactivity.

eydun:
Adding date/time to e-mail log message would make great sense.

dg:
Incredibly lazy community = no datetime in emails.

PR #498, opened 2021-07-12
BlueScreen: used scrubber for HTTP headers

dakujem:
Two sections of BlueScreen were previously not properly scrubbed:
- HTTP Reqeust / Headers
- CLI Request / Arguments

This includes, for example, the `Cookie` header, which can easily be used to hijack the session:
image

The Cookie is plaintext-legible in the `Headers` section, while the `$_COOKIE` section right below gets properly scrubbed:
image

This PR fixes the issue.

- bug fix
- BC break? no


dakujem:
Would you like me to create a second PR for the 2.x branch?

dg:
Thanks. I'll copy it to the 2.x branch.

Issue #499, opened 2021-08-03
BlueScreen callstack output can be looped by writing a exception with recursion previous property


janbarasek:
Version: all

### Bug Description

In Tracy, there is a way to use recursive exceptions to drop the entire capture process via repeatedly writing `previous`.

### Steps To Reproduce

How to kill your application.

Use this code:

```php
$a = new class extends \RuntimeException {
	public function setPrevious(\Throwable $e): void
	{
		$ref = new \ReflectionClass($this);
		$parent = $ref->getParentClass()->getParentClass();
		$previous = $parent->getProperty('previous');
		$previous->setAccessible(true);
		$previous->setValue($this, $e);
	}
};
$a->setPrevious($a);

// this line will kill your BlueScreen:
throw $a;
```

Dumped value is:

![Snímek obrazovky 2021-08-03 v 20 19 32](https://user-images.githubusercontent.com/4738758/128066547-a8ddb22f-bca1-42c8-afb3-8379924d0f77.png)

Note the recursion on the `previous` property, where Tracy recognizes it correctly. This logic must also be in the callstack statement.

### Expected Behavior

Stop callstack dumping in case of recursion, or at least limit the depth.

Thanks.

mabar:
Dump looks correct. Instance refers to itself and dump stopped due to recursion. It even shows the same object id `#369` to inform you it's the same instance.

I can't reproduce infinite loop killing application with following script - with Tracy enabled. But the same code in Nette presenter without Tracy enabled creates infinite loop. Problem is somewhere else, probably better use xdebug step debugging to find where the issue actually is.

```php
<?php declare(strict_types = 1);

require __DIR__ . '/../vendor/autoload.php';

\Tracy\Debugger::enable(true);

$a = new class extends \RuntimeException {
	public function setPrevious(\Throwable $e): void
	{
		$ref = new \ReflectionClass($this);
		$parent = $ref->getParentClass()->getParentClass();
		$previous = $parent->getProperty('previous');
		$previous->setAccessible(true);
		$previous->setValue($this, $e);
	}
};
$a->setPrevious($a);
throw $a;

```

PR #503, opened 2021-08-19
 Bugfix for bluescreen generated by AJAX

kuty:
Bluescreen method render() requires content.phtml directly without page.phtml. This is causing undefined variable $exceptions for exceptions in AJAX calls. Bugfix appeared after fixing issue #499.

- bug fix after changes done in #499 
- BC break? no
- doc PR: n/a

Variable $exceptions was not initiated when content.phtml was required directly without page.phtml in case of AJAX calls.


dg:
Fixed, thanks

Issue #504, opened 2021-08-20
Undefined variable $exceptions, did you mean $exception?


hubipe:
Version: 2.8.6

### Bug Description
Tracy debug panel show Notice about undefined variable $exceptions (see image below). When I downgraded Tracy to previous version (2.8.5), this notice was not fired (and instead the Tracy shown the actual error).

![image](https://user-images.githubusercontent.com/617880/130214630-e15bf075-92b0-4b78-b3d1-d70a99a31f83.png)

### Steps To Reproduce
I actually don't know,how to reproduce. I tried to make a simple demo, it handled the same type of error correctly. Maybe it has something to do with the actual request was an AJAX?

### Expected Behavior
Same error should be rendered as in the version 2.8.5:

![image](https://user-images.githubusercontent.com/617880/130213299-0f713c64-c168-4b78-971d-117e7576064b.png)

dg:
Fixed

PR #505, opened 2021-08-20
Undefined variable $exceptions in AJAX requests fix

hubipe:
- bug fix / new feature?   bug fix of #504
- BC break? no

This pull requests solve the #504 issue, where debugging window does not know the variable $exceptions in AJAX requests.


dg:
Thanks

PR #508, opened 2021-09-15
Compatibility with psr/log ^2.0 and ^3.0

mabar:
- new feature
- BC break? yes
- doc PR: not needed?

Minimum needed to be compatible with psr/log ^3.0. It should be safe to do on PHP >=7.2, any extending class just have to add return type too.


dg:
Thanks

PR #518, opened 2021-12-05
Allow to select error severity for Debugger::$scream

xificurk:
- new feature
- BC break? no
- doc PR: TBA if accepted

`Debugger::$scream` is currently simple on/off switch that unmutes all/none of the suppressed errors. This PR adds support for selecting which error severities should be unmuted.

This feature might be useful e.g. for tracking and logging of deprecation errors triggered in "muted" state by some of the popular libraries, see:
- https://github.com/doctrine/deprecations/blob/master/lib/Doctrine/Deprecations/Deprecation.php#L189
- https://github.com/symfony/deprecation-contracts/blob/main/function.php#L25

Setting `Debugger::$scream = E_USER_DEPRECATED` would cause all the deprecations to be tracked by Tracy, while still ignoring all other muted error types.

Second commit adds/improves the support for error severity configuration in `scream`, `strictMode`, `logSeverity` options of `TracyExtension`.

### DIFF
--- a/src/Bridges/Nette/TracyExtension.php
+++ b/src/Bridges/Nette/TracyExtension.php
@@ -11,6 +11,7 @@ namespace Tracy\Bridges\Nette;
 
 use Nette;
 use Nette\Schema\Expect;
+use Nette\Utils\Strings;
 use Tracy;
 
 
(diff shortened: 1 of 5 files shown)


dg:
Good feature.

I think nobody understands these notations like 'E_ALL & ~E_DEPRECATED & ~E_USER_WARNING', so it would probably be better to use (only) something like `strictMode: [E_DEPRECATED, E_USER_WARNING]` and `strictMode: allExcept: [E_DEPRECATED, E_USER_WARNING]`, what do you mean?

xificurk:
I don't know about the custom syntax...

On one hand, I really like the improved understandability for someone reading this in some config.

On the other hand, if I am the one that wants to set this up, I would have never guessed this custom syntax. Main PHP documentation and almost any tutorial for PHP error handling uses examples with the constants and bitwise operators. I would bet, if someone doesn't understand the meaning of `E_ALL & ~E_DEPRECATED`, then they probably also do not understand the meaning and differences between e.g. `E_USER_WARNING`, `E_WARNING` and other levels.

Another minor advantage of the standard expression with bitwise operators is that it's scalar, which can be much more easily loaded from ENV variable.

xificurk:
Thank you :+1:

PR #522, opened 2021-12-24
[2.5] Fix compatibility with PHP 8.1

jtojnar:
- bug fix
- BC break? no

Fixes a warning on PHP 8.1 (want to start using Tracy for [selfoss](https://github.com/fossar/selfoss), which still supports PHP 5.6).

### DIFF
--- a/src/Tracy/Helpers.php
+++ b/src/Tracy/Helpers.php
@@ -44,7 +44,7 @@ class Helpers
 	 * Returns link to editor.
 	 * @return string|null
 	 */
-	public static function editorUri($file, $line = null, $action = 'open', $search = null, $replace = null)
+	public static function editorUri($file, $line = null, $action = 'open', $search = '', $replace = '')
 	{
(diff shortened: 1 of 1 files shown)


dg:
I think there will be many more incompatibilities. And it's not necessary to use exactly version 2.5, isn't?

jtojnar:
2.6 is not compatible with PHP 5.6. But I encountered some more issues that are resolved in 2.9 so I will probably create a 2.9 fork and downgrade its PHP code to support 5.6.

dg:
Composer will install the version that is appropriate for the PHP you are using.

jtojnar:
Yes, that is why it installs 2.5.

dg:
For PHP 5.6 it installs Tracy 2.5
For PHP 8.1 it will install Tracy 2.8

Why would Tracy 2.5 support PHP 8.1?

jtojnar:
Because `composer.lock` can only contain a single version AFAIK.

PR #525, opened 2022-01-01
Debugger::timer: switch from microtime to hrtime

bckp:
hrtime is faster on virtual machines, due microtime use system call that need to change context from user to kernel (PHP7.3+)

- bug fix / new feature?   optimalization
- BC break? no
- doc PR: nette/docs#???  not needed


dg:
Thanks

PR #526, opened 2022-01-04
avoid installing /examples, to avoid treating it as source code

TomasVotruba:


dg:
I don't know if that's a good idea, because if someone downloads the zip, he' ll be surprised that it doesn't have the examples that he sees on github.

TomasVotruba:
Technically, that might happen. Yet nowadays it's standard to use composer to install dependencies. This file then makes sure it only containt src code.

Also, there is alredy 6 files/dirs that are ignored this way: https://github.com/nette/tracy/blob/79000ea5cb4fbcc148c99be6164a228e17882c1a/.gitattributes#L1-L6

mabar:
Alternatively it can be done in composer.json
```json
{
	"archive": {
		"exclude": ["/examples"]
	}
}
```

dg:
@TomasVotruba  Those ignored files are not expected by anyone, as opposed to the examples.

@mabar This seems to be related to the archive command https://getcomposer.org/doc/03-cli.md#archive

PR #540, opened 2022-04-19
Unlock session file before close

SendiMyrkr:
- bug fix
- BC break? no

There is a problem with locking when you use php to execute  bash script with background processes eg. https://stackoverflow.com/questions/71929617/php-8-1-not-release-flock-after-calling-exec

Unlocking before fclose seems to be a solution.


dg:
Did you run into any real locking problems with Tracy?

SendiMyrkr:
Yes, example in SO link is part of my code. When i run script with it while i have enadbled tracy than other request freeze on flock in Tracy. I did kill php with that few times because limit of fcgi runs out.

dg:
I understand. I'm glad I know that, because on Windows it behaves differently.

PR #542, opened 2022-05-04
Drop slash after `action` from editor uri pattern

simPod:
Jetbrains claims it's not supported https://youtrack.jetbrains.com/issue/IDEA-292787

- bug fix / new feature?   don't know rly
- BC break? yes (?)


dg:
It is expected by scripts https://github.com/nette/tracy/tree/master/tools/open-in-editor

simPod:
@dg adapted

dg:
Not merged because is BC break.

PR #544, opened 2022-05-23
Helpers: Autodetection of PhpStorm in Linux environment for Editor URI.

janbarasek:
- new feature
- BC break? yes

In a Linux environment or on a Mac, the default disk path to PhpStorm is always `/usr/local/bin/phpstorm`. This allows us to detect that PhpStorm is available and set the default path to the editor.

At the same time, on a Mac, all we have to do is set the URI and the click-through works immediately.

This feature was inspired by the implementation in PhpStan Pro.

Snímek obrazovky 2022-05-23 v 8 44 31


dg:
This is a BC break, then file creation stops working.

janbarasek:
@dg I fixed the editor URI string.

dg:
PhpStorm can do this?

dg:
Sure, you open an existing file. But it stops the file creation and editing features of open-editor.js. 

Or what it's good for `&search=%search&replace=%replace` inside `$editor = 'phpstorm://%action?file=%file&line=%line&search=%search&replace=%replace';`?

janbarasek:
Sorry for the slower reply, I needed to thoroughly test all scenarios.

In the end, it looks like you can only send an action to PhpStorm via URL to open a file. So what about replacing the URI to only do the `open` action? Even so, this will solve most of the common cases when working with dump data.

Issue #549, opened 2022-12-13
PHP 8.2: dynamic properties are deprecated


radimvaculik:
Version: 2.9.5

### Bug Description
With PHP 8.2, there is a deprecation error on dynamic properties.

### Code
https://github.com/nette/tracy/blob/master/src/Bridges/Nette/TracyExtension.php#L120

mabar:
I this case best solution would probably be to throw an exception in case logger is not instance of `Tracy\Logger`. You are using different `ILogger` implementation that does not support sending emails.

mabar:
Do you override the logger yourself? Or is it overriden by some logging extension?
I think I already solved that issue in both orisai/nette-monolog and contributte/monolog by preserving original logger (in case Tracy bridge is enabled)

radimvaculik:
@mabar I use `final class SentryDecoratedTracyLogger implements ILogger` and use `        $this->parentLogger = Debugger::getLogger();` to preserve Tracy logging.

mabar:
Oh, my solution is currently broken too.

It overrides logger returned by `Debugger::getLogger()` and disables autowiring for service `tracy.logger` so my `ILogger` implementation is autowired instead even if service `tracy.logger` is created before Monolog has chance to wire into Tracy.

But. Order matters. If Monolog extension is loaded first, Tracy sets options to incorrect logger.
My case would be fixed by replacing `Tracy\Debugger::getLogger()->fromEmail =` by `$this->getService('tracy.logger')->fromEmail =` in generated DIC

mabar:
@radimvaculik I fixed the issue in orisai/nette-monolog. You should be fine if you copy LazyTracyToPsrLogger changes to your SentryDecoratedTracyLogger https://github.com/orisai/nette-monolog/commit/6f39f961c583feb7061cbdb7435bfad16e6191cc#diff-34f516dc85e79726f910d1b0e989ea939cbf310611cf0507e69eee3b4a12a127R56-R110

josefsabl:
I have this problem as well, using Contributte/Logging. Obviously this code `Tracy\Debugger::getLogger()->mailer = ?;` should not rely on the property to maybe exist, rather there should be something like MailLogger interface having the setMailer function.

Adding @ to suppress the deprecation provides a quick fix.

slepic:
I have the very same problem in a custom implementation of `Tracy\ILogger`.

My current solution is to place a `public $mailer` property on my implementation and ignore it's value.

[...]

PR #550, opened 2023-01-02
Fix deprecation with custom logger that doesn't have public mailer property

josefsabl:
- bug fix (#549)
- BC break? no

### DIFF
--- a/src/Bridges/Nette/TracyExtension.php
+++ b/src/Bridges/Nette/TracyExtension.php
@@ -120,7 +120,7 @@ class TracyExtension extends Nette\DI\CompilerExtension
 		}
 
 		if ($this->config->netteMailer && $builder->getByType(Nette\Mail\IMailer::class)) {
-			$initialize->addBody($builder->formatPhp('Tracy\Debugger::getLogger()->mailer = ?;', [
+			$initialize->addBody($builder->formatPhp('if (($logger = Tracy\Debugger::getLogger()) instanceof Tracy\IMailLogger) { $logger->setMailer(?); }', [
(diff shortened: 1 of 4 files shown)


mabar:
Issue is not just about `mailer`, it includes also properties `fromEmail` and `emailSnooze`.
Fixing it by condition just continues to ignore error - user set a property, which is expected to do something, but in reality does nothing.
And fixing it by introducing new interface is a BC break making existing compatible loggers broken.

If a library replaces original logger, it is its responsibility to maintain compatibility.

Compatibility logger wrapper in its simplest version may look like this:

```php
use Tracy\ILogger;

class ReplacingLogger implements ILogger
{

	private ILogger $originalLogger;

	public function __construct(ILogger $originalLogger)
	{
		$this->originalLogger = $originalLogger;
	}
	
	public function log($value, $level = self::INFO)
	{
		$this->originalLogger->log($value, $level);
	}

	/**
	 * @param mixed $value
	 */
	public function __set(string $name, $value): void
	{
		$this->tracyOriginalLogger->$name = $value;
	}

	/**
	 * @return mixed
	 */
	public function __get(string $name)
	{
		return $this->tracyOriginalLogger->$name;
	}

	public function __isset(string $name): bool
	{
		return isset($this->tracyOriginalLogger->$name);
	}

	/**
	 * @param array<mixed> $arguments
	 * @return mixed
	 */
	public function __call(string $name, array $arguments)
	{
		return $this->tracyOriginalLogger->$name(...$arguments);
	}

}
```

dg:
Fixed dea9aa33fd2ab91e3123d90d83cffe00feb62f72

PR #551, opened 2023-01-03
TracyExtension: set mailer to logger service instead of staticly get logger

mabar:
- bug fix 
- BC break? hard to tell

Commit https://github.com/nette/tracy/commit/0b615627e1229d33abc016620c8239ebcb20654f introduces condition that breaks my logger. It connects Monolog and Tracy logger in a way it is possible to log from Tracy to Monolog and from Monolog to Tracy while maintaining configuration of these loggers.

Since condition was introduced, replacing Tracy\Logger causes `mailer` option to be ignored *if Monolog extension is loaded firsts*. Because at point where mailer is set, `Tracy\Debugger::getLogger()` already returns different instance.

broken

By changing `Tracy\Debugger::getLogger()` to `$this->getService('tracy.logger')` I can maintain logger compatibility, while keeping fix for #549.

fixed

Note: These options probably have the same problem as described in #549.
https://github.com/nette/tracy/blob/0b615627e1229d33abc016620c8239ebcb20654f/src/Bridges/Nette/TracyExtension.php#L104-L105

If we change these calls from static getLogger() to getService() we may as well check if service `tracy.logger` will create `Tracy\Logger` and if not and one of these options is set, warn user about incompatible logger. Because current condition fix is just ignoring incompatible options which should not be set, if logger does not support them.

Sorry for long description, it was just crazy complex to make loggers work perfectly fine together and I don't have any other way of fixing it.

### DIFF
--- a/src/Bridges/Nette/TracyExtension.php
+++ b/src/Bridges/Nette/TracyExtension.php
@@ -120,7 +120,7 @@ class TracyExtension extends Nette\DI\CompilerExtension
 		}
 
 		if ($this->config->netteMailer && $builder->getByType(Nette\Mail\IMailer::class)) {
-			$initialize->addBody($builder->formatPhp('Tracy\Debugger::getLogger()->mailer = ?;', [
+			$initialize->addBody($builder->formatPhp('if (($logger = Tracy\Debugger::getLogger()) instanceof Tracy\Logger) $logger->mailer = ?;', [
 				[new Nette\DI\Statement(Tracy\Bridges\Nette\MailSender::class, ['fromEmail' => $this->config->fromEmail]), 'send'],
 			]));
 		}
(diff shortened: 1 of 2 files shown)


dg:
I tried to solve it like this: 0d551ba7cafec4548ff5617bf358e3ece767cecc

It's just a hack for the patch version for now. In the next major release we'd fix it better.

mabar:
One more workaround on my side, but it seems to be working fine, thanks 
https://github.com/orisai/nette-monolog/pull/3/files#diff-34f516dc85e79726f910d1b0e989ea939cbf310611cf0507e69eee3b4a12a127

dg:
I must have misunderstood... So they'll use your PR instead.

dg:
Is this dea9aa33fd2ab91e3123d90d83cffe00feb62f72 good for you?

mabar:
Yes, that works without any workarounds, thanks.

PR #555, opened 2023-02-15
Debugger::barDump() can be temporarily disabled/enabled

forgie1:
To prevent overcrowded dump, barDump can be temporarily disabled/enabled on demand.

eg:

```

function one()
{
    ...
    \Tracy\BarDump();
    ...
}

\Tracy\Debugger::$barDumpOn = false;
foreach ($many as $row) {
    one();
}

\Tracy\Debugger::$barDumpOn = true;
one(); // only this needs to be barDumped
```


dg:
Thanks, but this can be done on the caller side with a simple `if` condition. I'd rather not add global mutable state to `Debugger` for this.

PR #556, opened 2023-03-08
New StreamLogger & MultiLogger

f3l1x:
- new feature: #281 #283 #280
- BC break: no

---

Ahoj. Zkusil jsem navazat na #281 a pripravil PR s 2 loggery dle komentare https://github.com/nette/tracy/issues/280#issuecomment-1446550503

Motivace: Mit moznost logovat na STDOUT v prostredi Docker a nebo i ciste v `php -S localhost:8000 .`. Zaroven mit moznost zaregistrovat vice loggeru zaroven. Aby slo logovat na STDOUT a mit treba i posilani logu do Sentry ([tady](https://github.com/contributte/sentry/blob/71e3b78156d84c2da85b7c8fdbf5c542e16b5b97/src/Tracy/MultiLogger.php) a [tady](https://github.com/contributte/sentry/blob/71e3b78156d84c2da85b7c8fdbf5c542e16b5b97/src/Tracy/SentryLogger.php)).


src/Tracy/Logger/StreamLogger.php
```diff
@@ -0,0 +1,66 @@
...
+	{
+		$this->path = $path;
+	}
+
+
+	public function log(mixed $message, string $level = self::INFO)
+	{
+		$line = $this->formatLogLine($message);
+
+		if (!@file_put_contents($this->path, $line . PHP_EOL)) { // @ is escalated to exception
```
JanTvrdik, line 30: you are missing `FILE_APPEND | LOCK_EX` flags

f3l1x: FILE_APPEND and LOCK_EX don't work with `php://stdout`. Tested. Any ideas?

src/Tracy/Logger/StreamLogger.php
```diff
@@ -0,0 +1,66 @@
...
+
+	public function __construct(string $path)
+	{
+		$this->path = $path;
+	}
+
+
+	public function log(mixed $message, string $level = self::INFO)
+	{
+		$line = $this->formatLogLine($message);
```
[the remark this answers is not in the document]
f3l1x: Yup, thanks.

src/Tracy/Logger/StreamLogger.php
JanTvrdik: Works for me. What is your platform and SAPI?

f3l1x:
```
➜ php -v
PHP 8.2.3 (cli) (built: Feb 15 2023 00:58:03) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.2.3, Copyright (c) Zend Technologies
    with Zend OPcache v8.2.3, Copyright (c), by Zend Technologies
```

OSX / M1 ultra 
SAPI: php

mabar: I would omit flags for php:// streams. These streams are already opened by php
https://www.php.net/manual/en/wrappers.php.php#wrappers.php.std

JanTvrdik: @f3l1x Try checking `stream_isatty(STDOUT)` and run PHP with and without redirecting stdout to file (`php test > file.txt`). It should be possible to lock stdout when it is redirected to file.

dg:
Did you name the StreamLogger so that it doesn't clash with FileLogger (when it also logs to a file). Since I want to put this modification in Tracy 3, there is no need to maintain compatibility and it's fine to modify FileLogger.

dg:
Hey, it's pretty much up to you, since you're the author of this proposal :-) 

Let me ask you a different question: what should be the difference between StreamLogger and FileLogger?

dg:
I'm a complete idiot, I thought there was a FileLogger in Tracy 🤦‍♂️

Ok, so it would probably be a good idea to remove the current Logger completely and replace it with a MultiLogger containing FileLogger, MailLogger and maybe something else.

The hypotetic StreamLogger should probably have a resource in the constructor, if it has a path to a file as string, it should be called FileLogger imho.

The question is how to incorporate the fact that an HTML file is generated and stored in directory. Should this be a task for the next logger?

JanTvrdik:
btw this is possible relevant https://github.com/nette/tracy/pull/281/commits/62058e2b7fb89da25affd940b66ca5d6ab801855

dg:
Yeah, BlueScreenLogger looks like another logger that should be there.

The question is how to pass that the html filename from BlueScreenLogger to the other loggers when using a MultiLogger?

JanTvrdik:
> The question is how to pass that the html filename from BlueScreenLogger to the other loggers when using a MultiLogger?

* 62058e2b7fb89da25affd940b66ca5d6ab801855 solves this by not having `BlueScreenLogger` as part of `MultiLogger`
* monolog-tracy-handler solves this by [pre-computing the bluescreen filename before any logger is called and storing it in monolog LogRecord object](https://github.com/mangoweb-backend/monolog-tracy-handler/blob/master/src/TracyProcessor.php#L22-L36).

None of those solution seem particularly elegant.

Issue #560, opened 2023-03-29
`Exposer` purges content of some implementations of `Ds\Collection`


dakur:
Version: 2.9.6

### Bug Description
When I have `Ds\Stack` in my application which is inside of some service registered to DIC and I want to render all items in the stack into custom tracy panel, they are all gone.

That's because `ContainerPanel` from `nette/di` dumps all services into text representation and part of dumping is a call to `Exposer`. `Exposer` has special `exposeDsCollection()` method for instances of `Ds\Collection` and runs `foreach` on the collection. But it doesn't count with the fact that some of `Ds\Collection` implementation have destructive behavior – such as [`Ds\Stack`](https://github.com/php-ds/polyfill/blob/v1.4.1/src/Stack.php#L139) ([ref](https://github.com/php-ds/polyfill/blob/v1.4.1/src/Stack.php#L9)) or [`Ds\Queue`](https://github.com/php-ds/polyfill/blob/v1.4.1/src/Queue.php#L135) ([ref](https://github.com/php-ds/polyfill/blob/v1.4.1/src/Queue.php#L9)).

### Steps To Reproduce
- install & run [this repo](https://github.com/dakur/tracy-exposer-ds-collection)
- see tracy panel under "logs" label

### Expected Behavior
To see:

```
- from presenter
- from panel
```

as `Logger::log()` is called in both places. But only `from panel` is rendered.

### Possible Solution
Call `$obj->toArray()` in `exposeDsCollection()` method. But I'm not sure if it doesn't have any [other consequences](https://github.com/php-ds/polyfill/blob/v1.4.1/src/Collection.php#L49), I'm not so experienced in semantics of various data structures and their conversions.

Other possibility can be to differentiate specific `Ds\Collection` implementations separately.

dg:
cc @enumag

enumag:
Hmm... interesting... maybe this could help?
https://github.com/nette/tracy/commit/4d8678b51a89c4fa0e6e777be58caaa3a41bb414

dakur:
@enumag Yep, that works too.

PR #561, opened 2023-03-30
clone Ds\Collection

enumag:
Fixes https://github.com/nette/tracy/issues/560


dg:
Can you add a test?

enumag:
@dg Test added. I also checked that the test fails without the clone. It does.

dg:
Thanks

PR #563, opened 2023-05-09
Pass Debugger::maxItems option to Dumper in Debugger:barDump()

Creeperface01:
I assume this is a bug and the option should be passed along with the others to allow setting max array items.


dg:
Thanks

PR #566, opened 2023-06-19
Server name from nette/http

janfejtek:
- new feature? yes
- BC break? I don't think so

When  using nette behind proxy or in containers, server name is localhost. When nette/http is used, it reads actual host from proxy headers - https://github.com/nette/http/blob/7a92a2a61aaf2ebb74df11bb4a6f56daf20a1ef6/src/Http/RequestFactory.php#L294

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,8 @@
 		"nette/tester": "^2.2",
 		"latte/latte": "^2.5",
 		"psr/log": "^1.0 || ^2.0 || ^3.0",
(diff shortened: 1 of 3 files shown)


dg:
A better way to do this is to allow the MailSender object to set $host directly, rather than passing the entire Request to it.

janfejtek:
Is this ok?

dg:
Great, thanks

PR #567, opened 2023-06-19
fwrite on closed resource fix

janfejtek:
- bug fix
- BC break? no

This is described here https://github.com/php/doc-en/issues/1581


dg:
Thanks

PR #568, opened 2023-07-02
Added checks if $_SERVER['argv'] is set

janfejtek:
- bug fix
- BC break? no

Variable argv does not have to be present. First check relates to Helper class, which does not add column into source when run from cli and argv is missing
https://github.com/nette/tracy/blob/209ac3040406e31ac4ab20a06cfcd0513a719b36/src/Tracy/Helpers.php#L173


dg:
Thanks

PR #587, opened 2024-10-27
Drop obsolete E_STRICT error level

xificurk:
- bug fix
- BC break? no

`E_STRICT` error level has not been used since PHP 8.0. PHP 8.4 started emitting deprecation warning on any usage of `E_STRICT` constant.

Tracy 2.10 requires min PHP 8.0, so this can be completely dropped.

See https://php.watch/versions/8.4/E_STRICT-deprecated


dg:
Thanks

PR #598, opened 2025-07-04
open-editor.js support for matching remote with wildcard

mabar:
- feature
- BC break no
- doc PR: Soon TM

This change makes it possible to use wildcard in remote path mapping,
e,g, `/var/www/dev*.example.com` will match  `/var/www/dev1.example.com` etc.
It is useful for remote debugging of apps with multiple dev instances

---

Note: order matters and I am not 100% the regex is correct, but it works on my machine


dg:
Thanks for the patch! After consideration, this feels like a fairly edge case — multiple dev instances on the same host can be handled by adding explicit mappings, and I'd rather keep open-editor.js minimal (it's a small JScript utility). Closing for now, but thanks for taking the time.

PR #600, opened 2025-08-12
Fix "Method ReflectionProperty::setAccessible() is deprecated since 8.5, as it has no effect"

staabm:
`setAccessible` is a no-op since 8.1 - see https://wiki.php.net/rfc/make-reflection-setaccessible-no-op


dg:
it is already removed a14cc6383b403fb931944ce62257710cda1c277b

PR #604, opened 2025-10-09
add missing meta tag viewport

mildabre:
This change adds the missing <meta name="viewport"> tag to the error page template to ensure proper scaling and layout on mobile devices. The update is non-invasive, does not affect desktop rendering, and improves usability on mobile.


dg:
Thanks! Added the viewport meta tag in 32996078 (on the v2.11 branch, will reach master with the next merge). Closing this PR since the change was applied directly to avoid the unrelated commits.

PR #605, opened 2025-10-09
Improve error 500 page layout for desktop and mobile

mildabre:
Description:

switched from absolute to fixed positioning for reliable viewport anchoring
applied width: 100vw and height: 100vh for full coverage
added missing background-color: white
smaller font sizes and margins for mobile devices

This update needs also add missing viewport defined in my previous PR #604


dg:
Thanks for the effort! The viewport meta tag from #604 has been applied, which already makes the page render well on mobile (tested on iPhone SE width).

I'm not taking the rest of the layout changes though:

- Switching to `position: fixed` + full-viewport white background would hide any partial output the app managed to render before the fatal error. I'd rather let users see whatever the app produced than cover it with a white overlay - even partial context is more valuable than none.
- With the viewport meta fixed, the existing font sizes and margins are readable on typical mobile widths (~375-400px), so the extra media queries aren't needed.
- The low-height landscape case is a rare edge case not worth the added CSS complexity.

Closing, but thanks again for spotting the mobile issue.

PR #609, opened 2025-12-22
Use `keysToHide` config for the DIC panel and debugger too

### LINKED ISSUE #608: `tracy.keysToHide` config in Nette config file not used for `dump()`/`bdump()`
Version: 2.11.0

### Bug Description
Tracy documentation has a section about [using Nette Framework configuration](https://tracy.nette.org/cs/configuring#toc-nette-framework).
It says that setting `tracy.keysToHide` will hide (sanitize) the values of the keys specified when a variable is dumped using `dump()`. To me it seems that's not the case, and that `tracy.keysToHide` is only used for the (red) bluescreen, and not for `dump()` or `bdump()`:
https://github.com/nette/tracy/blob/7543389d84c55e73b4fd2c4c9412da3aaeda3b18/src/Bridges/Nette/TracyExtension.php#L108

### Steps To Reproduce
Use the following config:
```neon
tracy:
	keysToHide: [password, pass, foobar]
```
Then ad the following somewhere in the project:
```php
dump(['password' => 'foo']);
bdump(['password' => 'foo']);
```
The array dumped to both the screen and the Tracy bar will have the password not hidden.

### Expected Behavior
I'd expect the password to be hidden, especially when the documentation says so.

### Possible Solution
I'm not sure, but the following works for me, as a proof of concept. Instead of setting `keysToHide` on the bluescreen object only, I set it on both.

Replace this:
https://github.com/nette/tracy/blob/7543389d84c55e73b4fd2c4c9412da3aaeda3b18/src/Bridges/Nette/TracyExtension.php#L107-L109
with the following code:
```php
				if ($key === 'keysToHide') {
					$initialize->addBody($builder->formatPhp('Tracy\Debugger::$keysToHide = ?;', Nette\DI\Helpers::filterArguments([$value])));
					$initialize->addBody($builder->formatPhp('array_push(Tracy\Debugger::getBlueScreen()->keysToHide, ... ?);', Nette\DI\Helpers::filterArguments([$value])));
					continue;
				}
				$tbl = [
					'fromEmail' => 'if ($logger instanceof Tracy\Logger) $logger->fromEmail = ?',

```

There's probably a better way but if there isn't, I'll happily prepare a pull request.

spaze:
- bug fix / new feature? new/possibly both
- BC break? no
- doc PR: will create one, I promise

Set `tracy.keysToHide` config for Debugger, DIC panel and `bdump` as well, not just for `BlueScreen`.

Will require https://github.com/nette/di/issues/327 because the `ContainerPanel::$keysToHide` property is added there, I'm not sure how to express it in `composer.json` (with `conflict`?)

I could also add a condition like:
```php
if (property_exists(Tracy\Debugger::class, 'keysToHide')) array_push(Nette\Bridges\DITracy\ContainerPanel::$keysToHide, ...$keysToHide)
```
to the generated code for `keysToHide` in `$tbl`, to avoid updating or adding any composer version constraints, which I think would be the best thing to do. Let me know what you think.

Close #608


spaze:
Thanks! I have updated https://github.com/nette/di/pull/328 which now uses the `Bluescreen` property to pass `keysToHide` to Dumper. I hope I got it right but if not please let me know.

PR #611, opened 2026-01-27
Describer: checked method existence

vrana:
- bug fix
- BC break? no
- doc PR: n/a

I couldn't pinpoint the exact configuration (my PHP version is 8.4.16) but under some conditions, an exception thrown inside an exception handler kills PHP-FPM in the most brutal way with no errors logged or printed anywhere. Debugging this was a nightmare. A simple check for function existence fixes the problem.


dg:
Hi Jakub, thanks for the patch! I'm a bit hesitant to merge a workaround for a bug I can't reproduce or fully understand — the  check might just mask a deeper issue rather than fix the root cause. Could you share a minimal repro, or at least a stack trace showing which call it is and what kind of object is involved? That would help me figure out what's actually going wrong. Thanks!

PR #612, opened 2026-02-18
Also find nonce in `script-src-elem`

adrianbj:
If you have onclick or other inline event handlers, you might put your nonce in `script-src-elem` rather than `script-src`. This will find either way.

- bug fix / new feature? new feature I suppose. 
- BC break? no


dg:
> This directive only specifies valid sources for inline script event handlers like onclick. It does not apply to other JavaScript sources that can trigger script execution, such as URLs loaded directly into [<script>](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/script)

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src-attr

adrianbj:
Hi @dg - yes, I understand that. Sorry, my bad - my note might have confused you - it should say `script-src-elem` like the title of the PR does, rather than `script-src-attr`

adrianbj:
Hi @dg - sorry to bug you, but just wanted to make sure you read my comment above - I messed up term in my initial commit note, but I think the code in the PR is correct and very helpful if you'd consider including it please.

adrianbj:
Hi @dg - sorry if I am missing something here, but I don't understand how https://github.com/nette/tracy/commit/ecb3dfb1a0e345c984912b4b26d1da41da1175a2 fixes this. 

If a site has a CSP and they set a nonce on `script-src-elem` but they don't set it on `script-src` then Tracy won't parse out the nonce and won't apply it to its own inline scripts preventing Tracy from loading.

It's quite legitimate to have a CSP with a nonce on `script-src-elem` but not on `script-src` so that you can add nonces to inline scripts but not scripts with a `src`.

Does that make sense, or am I not explaining properly?

adrianbj:
Just to follow up, I believe these three inline scripts will fail in the scenario I describe:
https://github.com/nette/tracy/blob/fea1ec595c66e6e4aed0a1c8bf3df850094d15bd/src/Tracy/Bar/dist/loader.phtml#L17

https://github.com/nette/tracy/blob/fea1ec595c66e6e4aed0a1c8bf3df850094d15bd/src/Tracy/Bar/Bar.php#L121

https://github.com/nette/tracy/blob/fea1ec595c66e6e4aed0a1c8bf3df850094d15bd/src/Tracy/Bar/assets/loader.latte#L17

adrianbj:
Thanks @dg for finally implementing this. I must admit I am a bit confused about the resistance and silence about your change of mind.

Anyway, hopefully others will find it useful, so thanks again.

dg:
Hi @adrianbj, sorry for the delayed response and thanks for both the PR and your patience. You were right about the `script-src-elem` case, and I appreciate you sticking with it. Thanks again!

adrianbj:
Thanks @dg  - appreciate all your hard work on Tracy!

PR #613, opened 2026-03-05
feat: Add lazy panel loading support

### LINKED ISSUE #530: Tracy panel lazy loading
Hello,

As a package developer, I like to provide debug information to developers via the Tracy panel. However, getting some data takes a long time and is not necessary to do in every request (for example, mapping all entities in Doctrine).

I'd like to introduce the concept of lazy loading of panels in Tracy, which will compile their content after the panel is clicked.

I've been thinking about the concept of builds, and lazy loading should only work for panels that work in isolation from the current request and can always be retrieved for their content. For example, the panel would be called as a DIC service and so on (I would be very happy for more ideas on how this could work).

Also, panels should natively be able to be set to not be rendered during ajax. This then delays the loading of the whole page, even though I don't need the panel most of the time.

Thank you.

jakduch:
## Summary

Implements lazy loading for debug bar panels (resolves #530). Panels registered with `lazy: true` have their `getPanel()` deferred to a shutdown function — the tab is rendered immediately, but the panel content is stored in the session and fetched via AJAX when the user clicks or hovers over the tab.

This is useful for panels that perform expensive operations (database profiling, API call logging, heavy data analysis) — they no longer slow down the page response time.

### Usage

```php
// Normal panel (eager) — rendered during the request
Debugger::getBar()->addPanel(new MyPanel, 'my-panel');

// Lazy panel — getPanel() is called in shutdown, content served via AJAX
Debugger::getBar()->addPanel(new MyExpensivePanel, 'my-panel', lazy: true);
```

### How it works

1. `Bar::addPanel()` accepts a new `lazy: bool = false` parameter
2. During `renderPanels()`, lazy panels skip `getPanel()` — only `getTab()` is called
3. After the response is sent, `renderLazyPanels()` calls `getPanel()` and stores content in session via `DeferredContent`
4. A new AJAX endpoint (`_tracy_bar=lazy-panel.{requestId}.{panelId}`) serves the panel content as JSON
5. On the client side, `bar.js` detects `data-tracy-lazy` panels and fetches content on first interaction

### Changes

- **`Bar.php`** — `addPanel()` with `lazy` param, `renderLazyPanels()` method, lazy-aware `renderPanels()`
- **`DeferredContent.php`** — New `lazy-panel` AJAX endpoint
- **`bar.js`** — `fetchLazyContent()` method, lazy-aware `Panel.init()`
- **`bar.phtml`** — Lazy panels get clickable `<a>` tabs
- **`panels.phtml`** — Lazy panels get `data-tracy-lazy="1"` attribute
- **`DevelopmentStrategy.php`** — Calls `renderLazyPanels()` after `render()`
- **`ProductionStrategy.php`** — Matching interface update

### Demo

A demo page is included at `examples/lazy-panels.php` with three panels: one eager and two lazy (simulating expensive operations with `usleep`).

### Backward compatibility

Fully backward compatible — the `lazy` parameter defaults to `false`, so existing `addPanel()` calls work unchanged. No changes to `IBarPanel` interface.


jakduch:
@dg thanks for the 👍!

I've merged current master into the branch and resolved the conflicts. It needed a bit of rework since the Bar frontend moved to Shadow DOM and the templates were reworked, so the lazy loading is now re-applied on top of the new architecture. The PR is conflict-free and mergeable again.

A couple of questions:
- Would you like me to keep it up to date with master going forward?
- Is landing lazy panel loading something you'd want to include in Tracy?
- Anything you'd like me to add or change to make it merge-ready (tests, docs, API tweaks)?

Happy to take it whatever direction works best for you.

